PR #1, opened 2021-04-26
Dynamic fees V1

jcape:
A simple proposal for start-time configurable fees.

[Rendered RFC](https://github.com/jcape/rfcs/blob/dynamic-fees-v1/text/0001-dynamic-fees-v1.md)


eranrund (approved):

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+MobileCoin launched with a hard-coded minimum fee of 0.01 MOB. When this constant was set, the value of MobileCoin was assumed to be slightly below 1 USD per coin, resulting in a per-transaction fee of under 0.01 USD (1c). In the intervening 6 months, the value per-coin increased to 50 USD, resulting in a per-transaction fee of 0.50 USD.
+
+This fee is too high to reasonably support the cannonical use-case---purchasing a cup of coffee---and we would like to lower the fee to $0.04 to bring it back in line with that use-case.
+
+However, this creates up a different problem. If the consensus system is roughly capable of completing 50tps (transactions per second), then the cost of conducting a denial of service attack is directly related to the size of the fees. At 0.50 USD/tx, consuming the entire consensus power of the network would cost over 2 MUSD/day.
```
sugargoat, line 18:
```suggestion
However, this creates a different problem. If the consensus system is roughly capable of completing 50tps (transactions per second), then the cost of conducting a denial of service attack is directly related to the size of the fees. At 0.50 USD/tx, consuming the entire consensus power of the network would cost over 2M USD/day.
```

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+At the desired 0.04 USD/tx, however, the cost of conducing a denial of service attack would be less than 175 kUSD/day. This may seem like a lot of money, but recall that at peak trading on FTX, 80 MMOB or 4 GUSD changed hands, which gives some sense of scale to those costs (i.e. they are too low).
```
sugargoat, line 20:
```suggestion
At the desired 0.04 USD/tx, however, the cost of conducing a denial of service attack would be less than 175k USD/day. This may seem like a lot of money, but recall that at peak trading on FTX, 80M MOB or 4G USD changed hands, which gives some sense of scale to those costs (i.e. they are too low).
```

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+The second issue is that changing the hard-coded fee requires going through the full release cycle: testnet deployment and testing, mainnet enclave signing, client updates to trust the new enclave, a grace period for client updates, then finally a coordinated mainnet deployment by operators.
+
+In our current environment, it takes at least a month to change the fee, and the fees are simultaneously too high and too low. This must change.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When a node operator starts the consensus service, it may optionally provide a configuration option or command-line argument overriding fee. The fee will be appended (as a string) to the existing to the key exchange "prologue" (a series of bytes that both sides agree upon in advance) when performing node-to-node attestation.
```
sugargoat, line 29:
```suggestion
When a node operator starts the consensus service, it may optionally provide a configuration option or command-line argument overriding fee. The fee will be appended (as a string) to the existing key exchange "prologue" (a series of bytes that both sides agree upon in advance) when performing node-to-node attestation.
```

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+If two nodes with different fees attempt to communicate, the result will be an `AeadError` during the handshake phase.
```
sugargoat, line 31: Could this error be more descriptive? e.g. `AeadError(FeeMismatch)`

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+The currently-configured fee should be made available to consensus clients directly by extending the `BlockchainAPI.GetLastBlockInfo` response. This gRPC call is already polled by thick clients to orient themselves with relation to the state of the network.
```
sugargoat, line 33: 👍

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The `consensus_service` application will be extended to add a new `--minimum-fee` command-line argument and `MC_MINIMUM_FEE` environment variable. These values will be merged (environment-variable, then command-line option). This value will contain a string indicating the quantity and SI-scale of the minimum fee. Fees with out units will be considered to be denominated in picoMOB. Therefore, all these values are equivilent fees:
```
sugargoat, line 38: "merged" meaning command-line option supersedes env var?

text/0001-dynamic-fees-v1.md
sugargoat: Hmm - I see why you indicated the SI unit modifier on the currency as opposed to the number - I was trying to avoid the confusion of "MUSD" = "MobileUSD" which could be a future stablecoin ticker.

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+This fee value will be injected into the attested key exchange by the enclave, by constructing a new UTF-8 string of the format: `"%s-%u"` using the existing `ResponderId`, and the u64 picoMOB-denominated representation of the new fee. This value is injected into the hashing scheme built onto the key exchange protocols in use.
+
+Because of this, two nodes with different fees will produce different encryption keys, which will lead to a failed decyption during the key exchange setup (specifically, the initiator will receive an `AeadError` when attempting to decrypt the initial message received from the responder).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The primary drawback to this approach is that it requires a restart of a quorum of the network in order to adjust to a new fee. In the intended common-case (low fees generally, higher fees in response to an active denial-of-service), this will entail restarting the network twice, and when a node is restarted, any pending transactions which did not make it into the current SCP slot will be discarded.
+
```
sugargoat, line 58: If it's not possible to be more descriptive in the `AeadError`, which it seems like it might not be, then I would add the challenge of debugging the attestation failure as a drawback.

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+This fee value will be injected into the attested key exchange by the enclave, by constructing a new UTF-8 string of the format: `"%s-%u"` using the existing `ResponderId`, and the u64 picoMOB-denominated representation of the new fee. This value is injected into the hashing scheme built onto the key exchange protocols in use.
+
+Because of this, two nodes with different fees will produce different encryption keys, which will lead to a failed decyption during the key exchange setup (specifically, the initiator will receive an `AeadError` when attempting to decrypt the initial message received from the responder).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The primary drawback to this approach is that it requires a restart of a quorum of the network in order to adjust to a new fee. In the intended common-case (low fees generally, higher fees in response to an active denial-of-service), this will entail restarting the network twice, and when a node is restarted, any pending transactions which did not make it into the current SCP slot will be discarded.
```
sugargoat, line 57: Worth mentioning that the restart protocols include letting the pending values drain while the nodes are set to not accept new transactions?

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+This is actually quite tricky to get right: obviously you want to immediately reject any transactions which do not meet the currently configured fee requirements, but it's entirely possible that your node is already queuing transactions when this comes in.
+
+In this scenario, a client submits a transaction, and it goes into a queue behind hundreds of other transactions. The fee is "correct" for the current block, but the transaction will not actually be entertained for acceptance until after the fee has been raised, and the fee is no longer valid.
+
+Chosing to simply accept that if any node has accepted a proposed transaction at a lower fee, that fee is (or would have been) valid for all nodes is an option, but this does make the ability to enforce fees dependent on the union of all SGX functionality, which is a change from our ad-hoc norms around the use of SGX as a privacy-enhancement technology, rather than a security-critical technology.
```
sugargoat, line 72:
```suggestion
Choosing to simply accept that if any node has accepted a proposed transaction at a lower fee, that fee is (or would have been) valid for all nodes is an option, but this does make the ability to enforce fees dependent on the union of all SGX functionality, which is a change from our ad-hoc norms around the use of SGX as a privacy-enhancement technology, rather than a security-critical technology.
```

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+The second option is to only check fees lazily, which will produce the scenario where a client could submit a transaction, and have it appear to "time out" later, because it was accepted into the queue at a low fee, then later judged to have an insufficient fee.
+
+The "restart for each change" system will have similar properties for users---when a node is restarted, any pending transactions it has are lost---but is dramatically simpler to implement, and does not require special thought about validity periods of a particular transactions' fee-paid status.
```
sugargoat, line 76: Another place to potentially clarify that the pending transactions are only lost in an emergency restart scenario, not if operators are able to follow restart protocols.

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,93 @@
...
+
+The primary rationale, then, is that this is "good enough" for right now, and we will follow this up with a "dynamic fees v2" that settles all questions properly under easier time constraints.
+
+# Prior art
+[prior-art]: #prior-art
+
+Unfortunately, due to the limited scope of this PR, prior art is going to be difficult to discuss. There is obviously a much larger conversation to be had about Etherium gas, Bitcoin transaction fees, etc., and how MobileCoin fees differ from them, but that honestly belongs in a future RFC. This is about doing the minimum viable to get us to a baseline fees, without compromising our ability to respond to denial of service attacks.
```
sugargoat, line 83:
```suggestion
Unfortunately, due to the limited scope of this PR, prior art is going to be difficult to discuss. There is obviously a much larger conversation to be had about Etherium gas, Bitcoin transaction fees, etc., and how MobileCoin fees differ from them, but that honestly belongs in a future RFC. This is about doing the minimum viable to get us to a baseline fee-adjustment strategy, without compromising our ability to respond to denial of service attacks.
```

text/0001-dynamic-fees-v1.md
jcape: Fair point.

text/0001-dynamic-fees-v1.md
jcape: Unfortunately no: the noise prologue is just a byte string that each side independently hashes together with the key exchange results to construct the symmetric key, and this proposal is saying "let's extend that to include the fee, as well as the [existing] responder ID", so now, if you get an `AeadError` when you are doing an attestation, it means "you and the node you are connecting do disagree about their responder ID (or there is a MiTM attack that has replaced a pubkey somewhere)", now would be updated to be: "disagree about their responder ID or minimum fee".

text/0001-dynamic-fees-v1.md
jcape: Right, I think MoUSD might be an elegant solution for the naming of future stuff ;-)

text/0001-dynamic-fees-v1.md
jcape: Yes, I'll update.

sugargoat (approved):

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,95 @@
...
+
+In this scenario, a client submits a transaction, and it goes into a queue behind hundreds of other transactions. The fee is "correct" for the current block, but the transaction will not actually be entertained for acceptance until after the fee has been raised, and the fee is no longer valid.
+
+Choosing to simply accept that if any node has accepted a proposed transaction at a lower fee, that fee is (or would have been) valid for all nodes is an option, but this does make the ability to enforce fees dependent on the union of all SGX functionality, which is a change from our ad-hoc norms around the use of SGX as a privacy-enhancement technology, rather than a security-critical technology.
+
+The second option is to only check fees lazily, which will produce the scenario where a client could submit a transaction, and have it appear to "time out" later, because it was accepted into the queue at a low fee, then later judged to have an insufficient fee.
```
UkoeHB, line 76: In practice:
1. there should be an expectation in the ecosystem about how much the 'lazy' minimum fee can vary over short timespans
2. the default fee used by clients can be some multiple of the minimum fee, above the potential variance over e.g. 100 blocks (current tombstone limit)

jcape:
Chat transcript:

@mfaulk  11:46

> The dynamic fees RFC looks solid to me, although it would be nice to not have to restart nodes in order to adjust the fee. Here's a simple suggestion that might help with that:
> The enclave only enforces fee constraints that don't need to be adjusted, e.g. "fee > 0".
> Each node maintains a minimum_fee parameter.
> When a client submits a transaction to a node, that node rejects the transaction if its fee is less than minimum_fee. Otherwise, it processes the transaction as normal.
> The idea here is that each node is a "gatekeeper" that enforces a fee minimum before a transaction gets into the consensus mechanism, but that once a transaction is in consensus, the fee value isn't important. In general, node operators will coordinate updating fees together, but nodes with different minimum_fee values would be able to coexist on the network together (which is nice for things like performing rolling updates). Since each node's minimum_fee value is not part of its enclave, it can be changed without restarting the node.
> The obvious downside to this is that a single node could set their minimum_fee very low, but that might not actually be a problem. That node would still have to validate every incoming transaction and propose the transactions for nomination, so the impact of this would be a fraction of transactions with unusually low fees getting accepted by consensus. This behavior would be visible to all nodes, and so could be handled at the "hey, your node is misconfigured" level.

@jcape  12:41
> I generally agree with that, but you could get to the same place by simply letting each node's minimum fee be independently configured.
12:41
> i.e. not do kex and let SCP tx validation handle it
12:42
> Like, my understanding of the transaction handling is that we do the is_well_formed() (read: is_valid()) test on all transactions we receive from clients, before we put them into the tx cache destined for SCP, and also do that on all tx received from peers.
12:43
> And the validation test includes the minimum fee.
12:44
> So just adjusting a particular node's fee and not telling anyone would have the same effect.
12:46
> Though there is one difference: an "authenticated" node (e.g. signal's node), lets you have a low fee for trusted clients, whereas unrestricted nodes can continue to charge high fees

@UkoeHB   12:47
> there is a difference between ‘ignoring a tx from client’ and ‘a tx is invalid’ - if you ignore a tx from client, you may still be ‘willing’ to accept that tx when a peer recommends it

@jcape   12:47
> Yes

@UkoeHB   12:47
> I think @mfaulk is recommending ‘ignore tx from client’

james  12:47
> Yes, he is (or that's my read of it). (edited) 

koe  12:49
> > let SCP tx validation handle it

> ah, this looked to me like adding a locally-defined ‘tx validity’ rule, which can break SCP (edited) 

@jcape   12:50
> Yes, that's my mis-reading of it.
> Implementing this directly means moving the fee validation out of `mc_transaction_core::validation::validate()` and into `client_tx_propose()` instead.

12:51

> There is a "race to the bottom" potential by excluding it from kex, though.

@UkoeHB   12:51
> race to the bottom?

@jcape   12:52
> e.g. [exchange] has a commercial incentive to keep their fees near zero

@UkoeHB   12:52
> oh I see, yeah it allows a perverse incentive (edited) 

@jcape   12:53
> I suppose that is sorted in the usual "OK, GTFO, bro" if a node operator is behaving badly.

@UkoeHB   12:54
> in terms of ‘short term solutions’, it may be simpler to implement

@jcape   12:54
> i.e. if someone leaves their fees low, and their node is a source of trash, other operators can just remove them

12:57

> Well, the RFC as-written is already mostly implemented (but not merged), changing it to this new version is pretty trivial, though---it's removing the hacks around adding the fee to the ResponderId, and moving the fee check to happen in a different place than it does currently.

text/0001-dynamic-fees-v1.md
UkoeHB: actually, both of those points apply to any solution, lazy or otherwise

text/0001-dynamic-fees-v1.md
```diff
@@ -0,0 +1,107 @@
...
+
+The "restart for each change" system will have similar properties for users as an emergency restart to introduce new fees---when a node is restarted without following the upgrade procedure, any pending transactions it has are lost---but is dramatically simpler to implement, and does not require special thought about validity periods of a particular transactions' fee-paid status.
+
+## Alt: Allow Divergent Fees
```
jayzalowitz, line 84: Divergent fee makes a lot of sense to me. For example I may be willing to conduct transaction processing when spot instances come up for sale on aws/azure/etc but not when that is not the case. This allows me to operate at 10% of the cost. I can leave a "pilot light" up at higher costs and then spin up unlimited clusters of more expensive software when that happens.
This makes the network almost impossible to ddos you'd have to be able to take down aws, azure, and gcloud simultaneously, and at that point why waste it on mob

jayzalowitz (commented):
Thoughts on cloud options wrt preventing any effective ddos

text/0001-dynamic-fees-v1.md
jcape: My instincts prefer divergent fees as well, but there's a bunch of gotchas that we'd have to talk our way through in order to get it to a place that feels comfortable---IMO it's a great candidate for the *next* "dynamic fees" RFC, where a fuller discussion is possible without the "OMG, tx costs $0.50" overhang :-)

jayzalowitz: So I can speak as a global expert when I say the following: Unless you standardize deployment in such a way that all the node operators have a couple of unneeded SPOF, everybody's infra is gonna run at a different cost, and whats more, at a certain repayment level, we can easily handle more load, its pretty much a demand economy in cloud nowadays.

(I will personally audit and make recommendations to node operators to run cheaper if you @ me on socials)

jayzalowitz: If a tx costs 50c to run, yall arent running all of the nodes in an efficient way. The aws minimum cost for an entire hour of time on a server that can run a (virtualized) SGX is 0.192000 ... are we really running one tx every 2.5 hours per node... seems unlikely.

kylefleming (approved):

jcape:
This has approval from consensus and the sdk and broad agreement, going to merge the RFC in 24h unless there's a big show-stopper issue in that time.

text/0001-dynamic-fees-v1.md
jcape: The "$0.50" refers to 0.01MOB fee at $50/MOB. Whatever the actual costs to validating node operators, it costs *users* $0.50 (well, $0.325 as of this writing) to conduct a transaction.

voloshyn (approved):

PR #3, opened 2021-08-13
Encrypted Memos

garbageslam:
A proposal to add encrypted memos to transaction outputs

[Rendered RFC](https://github.com/garbageslam/mcips/blob/encrypted-memos/text/0003-encrypted-memos.md)

PR #4, opened 2021-08-13
Recoverable transaction history

garbageslam:
A proposal to standardize several memo types to create a recoverable transaction history mechanism

This proposal is contingent on the encrypted-memos proposal being accepted, and this branch will be rebased on main if that happens.

[Rendered RFC](https://github.com/garbageslam/mcips/blob/recoverable-transaction-history/text/0004-recoverable-transaction-history.md)


eranrund (changes requested):

text/0003-recoverable-transaction-history.md
```diff
@@ -0,0 +1,328 @@
...
+which is read by the recipient, and some information in the change output memo, which is read by the sender.
+
+The normal output contains a "sender" memo which preserves information about the sender, for the recipient.
+The change output contains a "destination" memo which preserves information about the recipient, and the
+amount that was sent to the recipient, and the fee.
+
+## Sender memo
+
+The sender memo, with type bytes `0x0100`, includes a hash of the sender's address,
+and an HMAC value. This is used to inform the sender of who the recipient is in a verifiable way.
```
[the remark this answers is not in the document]
garbageslam: thanks

text/0002-encrypted-memos.md
```diff
@@ -0,0 +1,188 @@
+- Feature Name: Encrypted Memos
```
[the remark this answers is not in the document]
garbageslam: I guess I'm going to stage this PR on that PR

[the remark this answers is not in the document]
garbageslam: I can't do that unless I make the PRs from this repo instead of my fork, and I'd rather not do that, so I think we just have to ignore that file while reviewing this PR until that PR is merged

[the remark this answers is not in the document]
garbageslam: I guess I'm just going to make this a parallel PR, even though it doesn't make sense if the other one isn't merged

eranrund (approved):

garbageslam:
Feel free to join discussion in discord: https://discord.gg/TTQ5mE37

mfaulk (approved):

text/0004-recoverable-transaction-history.md
```diff
@@ -0,0 +1,351 @@
...
+- The domain separation string "mc-memo-mac"
+- The 32-byte public key of the transaction output to which this memo is attached
+- The memo type bytes
+- The first 28 bytes of the memo data (all of the data except the HMAC bytes).
+
+The HMAC output is truncated to 16 bytes.
+
+This HMAC text is computed in the same way for all category 0x01 memos.
+
+(See in code: https://github.com/mobilecoinfoundation/mobilecoin/pull/814/files#diff-681e5a738ac3ba20d8bcb29bcd66424b189075a4a2c1a27b9429b9da78aa7946R14)
```
jcape, line 180: I think MCIPs should follow stack-overflow rules, the answer should be self-contained instead of requiring a link.

(i.e. this should describe how the bytes are hashed, not link to a PR.)

text/0004-recoverable-transaction-history.md
```diff
@@ -0,0 +1,351 @@
...
+
+Here, the "total outlay" means the total amount that this transaction is deducting
+from the Sender's balance. It is the sum of the value of all non-change outputs,
+and the fee.
+
+The destination memo imposes some limits:
+* If the number of recipients exceeds 255 then the memo cannot be used. Since there are only 16 outputs
+  per transaction at most at time of writing, this is a safe assumption.
+* If there is more than one recipient, only one address hash can be recorded.
+  The client should choose one of them.
```
jcape, line 237: Any reason not to just use a single recipient here, and just say "destination memos can only be used with single-recipient txes", since it kind of falls apart if you don't?

Would that leave space for a MAC that would remove the need to keep the change address secret?

jcape (commented):
Mostly looks good, but I have some questions about the destination memo.

jcape (changes requested):
(marking as request changes for the algo description vs. PR link)

text/0004-recoverable-transaction-history.md
garbageslam: I already have a description of that, this is mainly just to help the readers who want to see a reference implementation

garbageslam: I can copy paste the hmac algo description from 0x0100 memo, but it's literally exactly the same? I can also just remove the code link

text/0004-recoverable-transaction-history.md
garbageslam: It would save us exactly one byte, and we are cannibalizing a byte from the fee to get that byte already

We have 12 unused bytes right now. If we use 16 bytes on a MAC here then we won't have any spare bytes to do something like, record payment request ID or refund ID or anything else which people may want later, so it seems more future proof if we use the subaddress validation idea that Kyle came up with

text/0004-recoverable-transaction-history.md
```diff
@@ -0,0 +1,347 @@
...
+| 0 - 16     | Sender's Address Hash |
+| 16 - 28    | Unused bytes          |
+| 28 - 44    | HMAC                  |
+
+The HMAC is computed using HMAC-SHA512. 
+(HMAC refers to [RFC 2104 HMAC](https://datatracker.ietf.org/doc/html/rfc2104).)
+(SHA512 refers to the 512 bit variant of [SHA-2](https://en.wikipedia.org/wiki/SHA-2).)
+
+The key to this HMAC is the 32-byte shared secret computed from Ristretto key exchange
+using the Sender's spend key, and the Recipient's view key. (For the sender, we multiply
```
jcape, line 164: Doesn't this mean the MAC key isn't uniform? Do we need to run this through an HKDF first?

jcape: (marking as resolved b/c we're using ristretto kex details elsewhere in similar situations, and also HMAC is going to run the key through SHA512 straight away anyways, HKDF just has an extra SHA512 round)

jcape (approved):

text/0004-recoverable-transaction-history.md
```diff
@@ -0,0 +1,347 @@
...
+belongs to a real person. This requires hashing about 2^128 addresses, which is similar in difficulty to brute-forcing
+our other cryptographic primitives.
+
+Additionally, finding a hash collision won't enable the attacker to bypass the HMAC check, it would just mean that the
+client has multiple hits for an address hash and must check HMAC against several addresses.
+
+Using a longer hash for the address hashes would require more space in the memo, and likely wouldn't help anything.
+
+One drawback of using an address hash is that if a later MCIP proposes to add new fields to the public address, the hash will change.
+This will require a more complicated versioning story for public addresses.
```
isis-mc, line 280: Obviously out-of-scope for this proposal, and I know I've mentioned this before several times, but this (sub)protocol versioning story is something we're going to want to address sooner rather than later.

text/0004-recoverable-transaction-history.md
```diff
@@ -0,0 +1,347 @@
...
+
+One drawback of requiring that the change subaddress is secret in an MCIP like this is that a client may concievably have already revealed
+the change subaddress to someone. However, it turns out that in practice, clients were already using the change subaddress this way, and
+we are just standardizing existing convention.
+
+One drawback of the Authenticated Sender memo design is that it doesn't have a full digital signature from the Sender, and instead uses
+this somewhat weaker HMAC-over-key-exchange mechanism. However, the HMAC mechanism results in a significantly shorter signature, and it
+achieves our deniability goal, which a full digital signature would not.
+
+One drawback of the Authenticated Sender With Payment Request Id memo design is that the pyament request id is limited to 8 bytes. However,
```
isis-mc, line 294: `s/pyament/payment/`

isis-mc (approved):
Some minor typos and comments, but looks great!

text/0004-recoverable-transaction-history.md
```diff
@@ -0,0 +1,347 @@
...
+In a mobile chat application, the use of these memos in payments permits the app to recover transaction
+history given only the private keys of the user, by fetching all of the user's TxOut's from Fog and then
+decoding and validating the memos.
+
+Some additional ancillary parts of the proposal:
+- Change subaddress is standardized to subaddress index 1.
+- Clients should always produce a change output even if the change is zero, in order to have a place to put
+  the destination memo.
+- A standard 16-byte hash of a public address is specified, which may be useful elsewhere.
+- A memo containing payment request id numbers is also introduced.
```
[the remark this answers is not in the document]
garbageslam: So the question for me is then, when those later MCIPs are merged, do we update this MCIP to refer to them?
I had thought it is not kosher to update an MCIP but perhaps an MCIP can be a living document? Or perhaps I'm misunderstanding your meaning?

text/0004-recoverable-transaction-history.md
garbageslam: I think it might not actually be such a big deal -- if we add a new optional field to public address, that doesn't change the hash of old public addresses, because of mc-crypto-digestible, and it's also compatible with protobuf.

If we add new mandatory fields, that will make it a lot more complicated

text/0004-recoverable-transaction-history.md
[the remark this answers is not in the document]
jcape: So it's obviously OK update prior MCIPs for typos, and I think practically keeping a registry is also fine---the changes to the earlier MCIP would be part of the PR, so "accepting"/merging the new PR would also update registry. This is something to work out in the other proposed scheme.

garbageslam:
This is now entering the final comment period

garbageslam:
We have delayed the final comment period on this proposal becaues MCIP #3 on which it depends has been delayed, due to additional comments. We will enter the final comment period again tomorrow if there are no more comments.

garbageslam:
This is entering the Final Comment Period again

PR #17, opened 2021-12-03
MCIP Draft - Ring Decoy Selection Clumps

UkoeHB:


garbageslam:
:+1: this is a good proposal, and I think it's only client-side changes to do this

PR #19, opened 2021-12-03
MCIP Draft - Simplified Encrypted Fog Hint

UkoeHB:


garbageslam:
I think the main issue with this is this part:

```
- If decryption succeeds, the decrypted message should be a compressed Ristretto point. Checking if it is a legitimate compressed Ristretto must be done in constant time.
```

Because, curve25519-dalek does not currently offer a way to do this, the decompress function returns `Option<RistrettoPoint>` which cannot be accessed in constant-time, and it has some early returns in the logic.

https://github.com/dalek-cryptography/curve25519-dalek/blob/6820effb4422a47f4fa68697f06def06a7567bfe/src/ristretto.rs#L254

We would need to talk to the developers to see if they are willing to introduce a constant-time decompression function to accommodate this. (@isis-mc )

PR #24, opened 2022-02-02
MCIP Draft - increase memo size

wjuan-mob:
A proposal to increase encrypted memos size.

[Rendered proposal](https://github.com/wjuan-mob/mcips/blob/main/text/0024-increase-memo-field-size.md)

### DIFF
--- a/text/0003-encrypted-memos.md
+++ b/text/0003-encrypted-memos.md
@@ -3,7 +3,7 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
-*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](). The changed byte tables are reflected there. 
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The changed byte tables are reflected there. 
 
 # Summary
 [summary]: #summary
--- a/text/0003-encrypted-memos.md
+++ b/text/0003-encrypted-memos.md
@@ -3,7 +3,7 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
-This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The changed byte tables are reflected there. 
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The changed byte tables are reflected there. 
 
 # Summary
 [summary]: #summary
@@ -73,6 +73,8 @@ After an interim period, a release will be made that makes memos mandatory.
 
 The release that makes memos mandatory will result in a new MRENCLAVE value for consensus, but not a new
 `BLOCK_VERSION` as it is not a breaking change to the ledger format.
+
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The changed byte tables are available [here](0024-increase-memo-field-size.md#this-is-an-update-to-the-values-from-3).
 
 The `e_memo` field, when present, is exactly 46 bytes long. Exactly 46 bytes of plaintext are selected when the TxOut is constructed, and then encrypted.
 
--- a/text/0003-encrypted-memos.md
+++ b/text/0003-encrypted-memos.md
@@ -3,6 +3,8 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
+This MPIC describes a memo which has had its size adjusted from 46 to 66 in a [later mcip](https://github.com/mobilecoinfoundation/mcips/pull/24) 
+
 # Summary
 [summary]: #summary
 
--- a/text/0003-encrypted-memos.md
+++ b/text/0003-encrypted-memos.md
@@ -3,7 +3,7 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
-*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The changed byte tables are reflected there. 
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](). The changed byte tables are reflected there. 
 
 # Summary
 [summary]: #summary
@@ -74,7 +74,7 @@ After an interim period, a release will be made that makes memos mandatory.
 The release that makes memos mandatory will result in a new MRENCLAVE value for consensus, but not a new
 `BLOCK_VERSION` as it is not a breaking change to the ledger format.
 
-*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The changed byte tables are available [here](0024-increase-memo-field-size.md#this-is-an-update-to-the-values-from-3).
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](). The changed byte tables are available [here](0024-increase-memo-field-size.md#this-is-an-update-to-the-values-from-3).
 
 The `e_memo` field, when present, is exactly 46 bytes long. Exactly 46 bytes of plaintext are selected when the TxOut is constructed, and then encrypted.
 
--- a/text/0003-encrypted-memos.md
+++ b/text/0003-encrypted-memos.md
@@ -3,7 +3,7 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
-This MPIC describes a memo which has had its size adjusted from 46 to 66 in a [later mcip](https://github.com/mobilecoinfoundation/mcips/pull/24) 
+This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24 
 
 # Summary
 [summary]: #summary
--- a/text/0003-encrypted-memos.md
+++ b/text/0003-encrypted-memos.md
@@ -74,7 +74,7 @@ After an interim period, a release will be made that makes memos mandatory.
 The release that makes memos mandatory will result in a new MRENCLAVE value for consensus, but not a new
 `BLOCK_VERSION` as it is not a breaking change to the ledger format.
 
-*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](). The changed byte tables are available [here](0024-increase-memo-field-size.md#this-is-an-update-to-the-values-from-3).
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. That [MCIP lists the changed byte tables](0024-increase-memo-field-size.md#this-is-an-update-to-the-values-from-3).
 
 The `e_memo` field, when present, is exactly 46 bytes long. Exactly 46 bytes of plaintext are selected when the TxOut is constructed, and then encrypted.
 
--- a/text/0003-encrypted-memos.md
+++ b/text/0003-encrypted-memos.md
@@ -3,7 +3,7 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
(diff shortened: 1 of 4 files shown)


garbageslam (approved):
this looks good to me, thank you!

eranrund (approved):

garbageslam (changes requested):
We should decide what our policy is here:

It might be a good idea to update the text of #3 and #4 MCIPs (and note that they were modified by #24) that was discussed in comments on those PR #4:

https://github.com/mobilecoinfoundation/mcips/pull/4#discussion_r698776600

Or we could just link to this MCIP from those MCIPs, noting that they were updated I guess?

jcape:
I think this PR should modify the previous MCIP to add a note to the top (below the header) indicating that the size has been adjusted in this MCIP.

text/0003-encrypted-memos.md
```diff
@@ -3,6 +3,8 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
+This MPIC describes a memo which has had its size adjusted from 46 to 66 in a [later mcip](https://github.com/mobilecoinfoundation/mcips/pull/24)
```
jcape, line 6:
```suggestion
This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24 
```

text/0024-increase-memo-field-size.md
```diff
@@ -0,0 +1,103 @@
...
+
+Increasing memo size to the limit of allowed size for the EtxOutRecord for this bucket level would leave no room for error in requirements for the tokens and other features which are in development. Therefore it is desireable to leave a small fudge factor.
+
+## Alt: Move to a variable length memo size
+
+This would go against our privacy goals because variable length memo fields would leak meta data and also cause issues when using ORAM.
+
+## Rationale
+
+This is a minimal change that aligns and standardizes our number of items per bucket to 8, and reserves as much space as possible for future uses that we didn't anticipate, before starting to impact the performance of Fog / the cost of running Fog given that we anticipate additional space per ETxOutRecord being used for Tokens.
```
garbageslam, line 86: I might add: 64 bytes is a good number because now the memo is big enough to hold a Schnorr-style digital signature (which in some variants is 64 bytes and in some variants is 48 bytes, for 128 bit security)

garbageslam:
I think we should put the link to this in 0004 mcip and not only the 0003 mcip

jcape (commented):
This generally looks OK to me, I'd like to have a bit more about the potential use cases here--e.g. I wasn't convinced in chat until it was brought up that this made the field large enough to store a signature.

jcape:
> I think we should put the link to this in 0004 mcip and not only the 0003 mcip

Yeah, this MCIP should either update the original 0004 MCIP (specifically the byte tables), or link to the section in this MCIP where those are redefined.

jcape:
Looking at the 0004 tables, we probably want to move the HMAC to the end so we can calculate it in a reasonable way :-)

wjuan-mob:
> Looking at the 0004 tables, we probably want to move the HMAC to the end so we can calculate it in a reasonable way :-)

Sorry for my ignorance. Do you mind elaborating on this? What would make that more reasonable?

garbageslam:
the hmac is computed by taking all the other data in the memo as one of the arguments, if that data is discontinuous then it has to be copied so that it will become contiguous. but that's needlessly complicated, the hmac could just be the last item in the buffer.

text/0024-increase-memo-field-size.md
```diff
@@ -0,0 +1,103 @@
...
+[motivation]: #motivation
+
+Current space in the bucket outside of that being used for the memo does not leave enough additional space for tokens and other upcoming features without impacting fog performance by moving memory size into the next bucket. Therefore, since we anticipate this performance hit, we would like to expand the memo appropriately so that we uniformly standardize to the next bucket while leaving space for those features. This will also provide additional space in the memo for storing interesting features for the user such as elliptic curve points, as desired.
+
+We determined that 66 bytes was the largest we could make the memo with minimal negative impact on Fog, and that this was also more than adequate for use-cases that were considered most interesting. 
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This is an expansion of the memo data filed from the [encrypted memo feature](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0003-encrypted-memos.md#alt-use-a-longer-memo-size)
```
remoun, line 21: Let's make this a relative link? And elsewhere in this file.

```suggestion
This is an expansion of the memo data filed from the [encrypted memo feature](0003-encrypted-memos.md#alt-use-a-longer-memo-size)
```

text/0024-increase-memo-field-size.md
```diff
@@ -0,0 +1,103 @@
+- Feature Name: (`Recalculate appropriate memo size`)
+- Start Date: (2022-02-02)
+- MCIPS PR: [mobilecoinfoundation/mcips#0024](https://github.com/mobilecoinfoundation/mcips/pull/0024)
```
remoun, line 3: These extra zeroes gave me pause. How about we remove them?

```suggestion
- MCIPS PR: [mobilecoinfoundation/mcips#24](https://github.com/mobilecoinfoundation/mcips/pull/24)
```

text/0024-increase-memo-field-size.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Rationale
+
+This is a minimal change that aligns and standardizes our number of items per bucket to 8, and reserves as much space as possible for future uses that we didn't anticipate, before starting to impact the performance of Fog / the cost of running Fog given that we anticipate additional space per ETxOutRecord being used for Tokens. This additional space provides flexibility for any requirements from the SDK team for the memo field. More specifically, by increasing the memo size to 64 bytes the memo is big enough to hold a Schnorr-style digital signature (which in some variants is 64 bytes and in some variants is 48 bytes, for 128 bit security). 
+
+# Prior art
+[prior-art]: #prior-art
+
+This is an expansion of the existing Encrypted Memo's proposal in line with [anticipated future work](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0003-encrypted-memos.md#future-possibilities), and specifically referencing to the [alternative larger memo size proposal](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0003-encrypted-memos.md#alt-use-a-longer-memo-size).
```
remoun, line 91: Ditto (relative links)
```suggestion
This is an expansion of the existing Encrypted Memo's proposal in line with [anticipated future work](0003-encrypted-memos.md#future-possibilities), and specifically referencing to the [alternative larger memo size proposal](0003-encrypted-memos.md#alt-use-a-longer-memo-size). 
```

text/0003-encrypted-memos.md
```diff
@@ -3,6 +3,8 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
+This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24
```
remoun, line 6: Consider linking to the MCIP

```suggestion
This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](0024-increase-memo-field-size.md)
```

text/0004-recoverable-transaction-history.md
```diff
@@ -3,6 +3,8 @@
 - RFC PR: [mobilecoinfoundation/mcips#4](https://github.com/mobilecoinfoundation/mcips/pull/4)
 - MobileCoin Epic: None
 
+This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The byte tables have been updated to reflect this change.
```
remoun, line 6: Ditto (link)

text/0004-recoverable-transaction-history.md
```diff
@@ -148,13 +150,14 @@ Three new memo types are specified:
 
 ## 0x0100 Authenticated Sender Memo
 
-The 44-byte memo data is laid out as follows:
+The 64-byte memo data is laid out as follows:
```
remoun, line 153: It seems odd to adjust the older MCIP to the newest values. That makes it necessary to look at git history to grok the protocol evolution. I believe the link at the top is sufficient, maybe with a similar note here, but not changing this table -- instead duplicate it in the new MCIP.

text/0003-encrypted-memos.md
wjuan-mob: @jcape I'm actually not sure why you asked me to remove the link. Do you mind clarifying this?

text/0004-recoverable-transaction-history.md
wjuan-mob: @jcape Could you weigh in on this also? Do you have a preference?

text/0003-encrypted-memos.md
wjuan-mob: This appears unnecessary since the #24 seems to automatically link it.

text/0004-recoverable-transaction-history.md
wjuan-mob: The standard in rust appears to be to add a link to the update but not actually change the older file except to note that it has been superceded: https://github.com/rust-lang/rfcs/pull/2050 . I think remoun's suggestion is more consistent with this.

text/0003-encrypted-memos.md
```diff
@@ -3,6 +3,8 @@
 - RFC PR: [mobilecoinfoundation/mcips#3](https://github.com/mobilecoinfoundation/mcips/pull/3)
 - MobileCoin Epic: None
 
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](). The changed byte tables are reflected there.
```
remoun, line 6: Nit: Remove empty link
```suggestion
*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. The changed byte tables are reflected there. 
```

text/0003-encrypted-memos.md
```diff
@@ -72,6 +74,8 @@ After an interim period, a release will be made that makes memos mandatory.
 The release that makes memos mandatory will result in a new MRENCLAVE value for consensus, but not a new
 `BLOCK_VERSION` as it is not a breaking change to the ledger format.
 
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](). The changed byte tables are available [here](0024-increase-memo-field-size.md#this-is-an-update-to-the-values-from-3).
```
remoun, line 77:
```suggestion
*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in MCIP #24. That [MCIP lists the changed byte tables](0024-increase-memo-field-size.md#this-is-an-update-to-the-values-from-3).
```

text/0004-recoverable-transaction-history.md
```diff
@@ -3,6 +3,8 @@
 - RFC PR: [mobilecoinfoundation/mcips#4](https://github.com/mobilecoinfoundation/mcips/pull/4)
 - MobileCoin Epic: None
 
+*Note:* This MCIP describes a memo which has had its size adjusted from 46 to 66 in [MCIP #24](). The changed byte tables are reflected there.
```
remoun, line 6: Ditto (missing link)

text/0024-increase-memo-field-size.md
```diff
@@ -0,0 +1,176 @@
...
+
+A 66-byte encrypted-memo is computed by AES-256 in counter mode:
+
+```
+e_memo = AES256Ctr(memo_payload, memo_aes_key, memo_aes_nonce)
+```
+
+The fog-view enclave uses an Oblivious Map to privately map from fog-search-keys (16 byte outputs from a KexRng) to ETxOutRecords. The hashmap divides each bucket into as many KeySize + ValueSize pairs it can, which are consecutive ranges of bytes with no padding at all. These sizes are fixed at compile-time for the enclave.
+
+As seen in the [original proposal](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0003-encrypted-memos.md#alt-use-a-longer-memo-size) for encrypted memos, the number of items that we can fit into a bucket is `2048 / (16 + 1 + sizeof(ETxOutRecord))` rounded down to the nearest integer.
```
remoun, line 71: Ditto (relative)
```suggestion
As seen in the [original proposal](0003-encrypted-memos.md#alt-use-a-longer-memo-size) for encrypted memos, the number of items that we can fit into a bucket is `2048 / (16 + 1 + sizeof(ETxOutRecord))` rounded down to the nearest integer.
```

garbageslam (approved):
thank you!

the-real-adammork (approved):

mfaulk (approved):

wjuan-mob:
It's been 1 business day since Final comment period. Merging.

voloshyn (commented):
👍

PR #25, opened 2022-02-05
Confidential Token Ids

sugargoat:
A proposal for confidential token IDs to support multiple asset types.

[Rendered Proposal](https://github.com/sugargoat/mcips/blob/confidential-token-ids/text/0025-confidential-token-ids.md)


text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+A transaction specifies the TokenType for the entire transaction, and includes a "Proof of Opening" that all `token_ids` in the inputs and outputs are the same. This prevents spending one TokenType as a member of another TokenType. The TransactionBuilder now takes a `token_id` as an argument in its initializer. The TokenType is set in the clear on the TxPrefix for a transaction, which is only accessed in the enclave, after the payload is delivered via an encrypted, attested channel. This is similar to the TxInputs, and thus has a similar threat model for confidentiality. See [Confidentiality Analysis](#confidentiality-analysis) for an assessment of risk.
+
+Clients must choose how to expose multiple asset types to their users. There may be clients who wish to only expose a subset of asset types, or clients who wish to provide balances and transactionality in all asset types. 
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
```
garbageslam, line 29: I would consider adding a section here calling out explicitly how `Amount` has changed:

Previously, an amount of picomob in a TxOut is represented by a Pedersen commitment to that number, as well as a "masked_value", which is created by XORing the number of picomob with a pseudorandom mask (derived by hashing a shared secret). The purpose of the `masked_value` is that when the recipient receives a TxOut, they can unmask this value, then reconstruct the Pedersen commitment from the number, to confirm successful recovery of the value.

After this change:
* A `masked_token_id` is added, which, similarly to masked_value, provides a way for the recipient to learn the token id (and then subsequently confirm it).
* The pedersen commitment to the value now is made using a basepoint `H_i` which depends on the token id. (The basepoint for the blinding factor is the same as before regardless of the token id.)
* Thus, the recipient can attempt to scan a TxOut by first constructing the shared secret, then unmasking the masked_amount and masked_token_id, and then reconstructing the Pedersen commitment. If this reconstruction is successful, then the recipient knows the amount and the token id, similarly as before the token ids.

### Pedersen generators

The requirements for the pedersen generators are:
* The generator used for MOB is the same as before, so `H_0` is a known value.
* The set of all generators used for token ids should be cryptographically "orthogonal", meaning that, it should be intractable for anyone to find a linear relationship between them.
* The function mapping a token id `i` to the generator `H_i` must be implemented in constant-time.

The established value for `H_0` in mobilecoin is to take the `RISTRETTO_BASEPOINT` bytes, hash them using blake2b, and then hash this to the ristretto curve using the elligator map exposed by curve25519-dalek.
We propose to compute `H_i` by converting `i` to little endian bytes, and XOR'ing these bytes over the first four bytes of the `RISTRETTO_BASEPOINT` bytes before hashing them with blake2b, because this is a simple way to meet these requirements.

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+A simple proposal is that the enclave discontinue fee aggregation, so that the number of fee outputs scale linearly with the number of transactions in a block (not output TXOs). For example, a block with 3 transactions includes 3 fee outputs. Because each transaction can mint up to 16 output TXOs, there may be a variable number of outputs in the block. 
+
+Note that the untrusted portion of the node software knows how many transactions are in each block, and this is not considered statistically relevant information with respect to confidentiality. In other words, the current fee aggregation behavior is not meant as a confidentiality measure, purely as a space-saving measure.
+
+### Fee Ratio and Transaction Sorting
```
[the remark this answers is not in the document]
garbageslam: I think I would suggest to
* the number that is passed to untrusted is the `priority` of the Tx
* we specify that priority is computed by `fee / minimum fee`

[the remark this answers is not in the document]
garbageslam: I think I would suggest to
* the number that is passed to untrusted is the `priority` of the Tx
* we specify that priority is computed by `fee / minimum fee`

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Upgrade Plan
+
+As this is a substantial change to the ledger format and transaction protocol, we need to be thoughtful about what rollout looks like for this change to land with minimal impact on existing clients and maximum backward compatibility.
+
+The rollout proposal is the following:
```
garbageslam, line 222: I want to suggest that we make a second MCIP for the roll out proposal -- I actually think we may need to rethink the plan for rolling out recoverable transaction history to use a similar mechanism as this

garbageslam: https://github.com/mobilecoinfoundation/mcips/pull/26

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+2. Clients update to:
+    - at a minimum:
+        - construct transactions with the Proof of Opening and the `masked_token_id` in output TXOs, not allowing any new outputs to use the default empty bytes array for the `masked_token_id` field.
+        - default to `token_id = 0` when processing TXOs without a `masked_token_id` field
+        - obtain the `minimum_fee` for the default token (MOB)
+    - optionally expose multiple asset types by:
+        - obtain the `minimum fee` for all assets
+        - support constructing transactions with arbitrary `token_id`
+        - display balance for some set of `token_ids`
+3. The consensus validator nodes are deployed and allow for both new style and old style transactions to be constructed, validated, and externalized to the ledger for some grace period. However, no new outputs can be spent in old-style transactions, and new outputs must have an explicit `token_id` even if it's an old-style transaction
```
garbageslam, line 234: so unfortunately, this leads to breakage if the new client send a transaction to someone with an old client, because the old client can't spend the money.

I think we should do James' idea which I tried to write up here: https://github.com/mobilecoinfoundation/mcips/pull/26
and I think we should also plan to roll out the encrypted memos that way

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+2. The non-interactive verifier chooses challenge, `e`, by hashing the commitment and the value `d` from the previous step.
+
+    ```
+    e = Blake2B(proof_tag_e | commitment | d)
+    ```
+
+3. The prover computes the following, and includes these values in the Proof of Opening:
+
+    ```
+    u = y + e * token_id
```
UkoeHB, line 98: Is this supposed to be `u = y + e * amount`?

Don't you need a range proof on `commitment`? Am I wrong in thinking range proofs qualify as 'proofs of opening'?

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+    - Note also, the blinding factor base point does not need to be unique per TokenType, so we use the same blinding factor base point for all TokenTypes.
+    - Note finally, the Proof of Opening requires all commitments in a transaction were generated using the same group generator, base point `H_i`. If the commitments in input and/or output TXOs in a transaction are constructed with different group generators, and the sender is able to unwrap the sum as specified in the protocol below, it implies that the sender knows a nontrivial linear relationship among the `H_i`, which contradicts the assumption that the `H_i` are orthogonal. 
+
+The non-interactive zero knowledge Proof of Opening protocol is the following: (See [Zero Knowledge Proofs and Commitment Schemes, Page 27](https://www.cs.purdue.edu/homes/ninghui/courses/555_Spring12/handouts/555_Spring12_topic23.pdf))
+
+
+1. The prover establishes a value, `d`, calculated by hashing their secret values from the commitment (the `token_id` and the `blinding_factor`). Implementation note, we may use [Merlin](https://docs.rs/merlin/latest/merlin/) rather than Blake2B.
+
+    ```
+    y = Blake2B(proof_tag_t | token_id)
```
UkoeHB, line 84: If `token_id` is only 32 bits, this is not secure (`u` can be trivially guess-and-checked).

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+    
+This protocol is non-interactive because the prover anticipates the verifier's `e` value, so requires no interaction from the verifier.
+
+Thus, the Proof of Knowledge could be represented, for example, with:
+
+```
+struct ProofOfKnowledge {
+
+    /// Prover value established from hashes of commitment and blinding factor secrets
+    CompressedRistretto d,
```
UkoeHB, line 119: Capital letters for curve points make it much easier to disambiguate curve points from curve scalars.

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+    e = Blake2B(proof_tag_e | commitment | proof_of_knowledge.d)
+    proof_of_knowledge.d + (commitment * e) == G * proof_of_knowledge.u + H * proof_of_knowledge.v
+}
+```
+
+### Fee Outputs
+
+Currently, the enclave aggregates the fees for multiple transactions in a block, in order to mint a single fee output for the block and conserve space. To support multiple confidential TokenTypes, when minting fee outputs, the enclave must create multiple fee outputs if multiple TokenTypes were used in the block. A consideration here is that we would like for an observer not to discern whether a block contains transactions of multiple asset types by statistically analyzing the number of outputs to derive information about the number of fee outputs. 
+
+A simple proposal is that the enclave discontinue fee aggregation, so that the number of fee outputs scale linearly with the number of transactions in a block (not output TXOs). For example, a block with 3 transactions includes 3 fee outputs. Because each transaction can mint up to 16 output TXOs, there may be a variable number of outputs in the block.
```
UkoeHB, line 144: This would amplify the problem I discussed in the [fee epochs MCIP](https://github.com/mobilecoinfoundation/mcips/pull/10).

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Fee Ratio and Transaction Sorting
+
+Currently the [WellFormedTxContext](https://github.com/mobilecoinfoundation/mobilecoin/blob/a092039a4a5cc7f5e3e16eeadd0b2bc3a12667ae/consensus/enclave/api/src/lib.rs#L50) contains the fee, which must be used by untrusted to [sort transactions](https://github.com/mobilecoinfoundation/mobilecoin/blob/a092039a4a5cc7f5e3e16eeadd0b2bc3a12667ae/consensus/enclave/api/src/lib.rs#L131) when [combining transactions](https://github.com/mobilecoinfoundation/mobilecoin/blob/a092039a4a5cc7f5e3e16eeadd0b2bc3a12667ae/consensus/service/src/tx_manager/untrusted_interfaces.rs#L40) for consideration by the enclave, prioritizing higher fees for inclusion in the block during periods of network congestion. In order to keep the `token_id` confidential, we must provide a mechanism for untrusted to accurately sort transactions before asking the enclave to do work on the transactions, without revealing the `token_id`, and at the same time, allowing for different asset types to express differing minimum fee requirements.
```
UkoeHB, line 150: How can users decide what fee to use if fees are hidden (that seems to be your goal here)?

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+    ```
+    unmasked_token_id = masked_token_id ^ Blake2b(token_id_tag | shared_secret)
+    ```
+
+### Proof of Opening: Proof of Knowledge of Uniform Token ID in the Transaction
+
+We would like to be able to prove that all transaction inputs and outputs are using the same `token_ids`. We do this using a non-interactive, zero knowledge Proof of Opening that the sender has the capability to unwrap the sum of output commitments in the transaction, in conjunction with the homomorphic encryption property of [Pedersen Commitments](https://www.cs.cornell.edu/courses/cs754/2001fa/129.PDF), namely that addition on the commitments preserves the additive relationship of the pre-committed values, as long as they are using the same group generator.
+
+- To prove each TXO's commitment to the `amount` value, we use a Pedersen commitment, consisting of a group generator (`H_i`) raised to the value of the `amount`, with a blinding factor to prevent brute-forcing the full range of values (`2^64`). We compute the blinding factor as the blinding group generator raised to a hash of the shared secret for the TXO.
```
UkoeHB, line 67: Honestly I am sort of confused by this 'Proof of Opening' approach. You need a range proof on all output amount commitments, so I'm pretty sure it's sufficient to say: 1) all output amount commitments are range-proofed on `H_i`, 2) sum(inputs) == sum(outputs). The range proofs on outputs will show that `sum(outputs) = _ G + _ H_i` (for some unimportant values `_`), which implies `sum(inputs) == _ G + _ H_i` (all inputs were previously range proofed, which means it isn't possible for a sum of inputs to overflow - the only way you could have `sum(inputs) == _ G + _ H_i + _ H_n`).

garbageslam: Hmm, I think you are right, I think the thing being proved by the proof of opening is actually already implied by the range proofs being over `H_i`. I'm trying to figure out if there is some subtle loss in security parameter around that, but I don't see that there is.

text/0000-confidential-token-ids.md
garbageslam: Yeah this is not what I did in the draft PR, my draft code looks like this:

```

        // Compute prover secrets y and s, and the commitment based on these.
        //
        // These are unpredictable assuming that the blinding factor is unpredictable
        // and that Merlin adequately implements Fiat Shamir.
        let (y, s) = {
            let mut transcript =
                MerlinTranscript::new(PROOF_OF_OPENING_SECRET_DOMAIN_TAG.as_bytes());
            value.append_to_transcript(b"x", &mut transcript);
            blinding.append_to_transcript(b"r", &mut transcript);

            let mut y_bytes = [0u8; 32];
            let mut s_bytes = [0u8; 32];

            transcript.challenge_bytes(b"y", &mut y_bytes);
            transcript.challenge_bytes(b"s", &mut s_bytes);

            (Scalar::from_bits(y_bytes), Scalar::from_bits(s_bytes))
        };
```

so that the blinding factor feeds into both prover secrets and can't be brute forced

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The MobileCoin ledger protocol currently supports one token type, MOB. We would like to allow the addition of assets to the ledger without revealing which assets were involved in any transaction. The transaction confidentiality and integrity guarantees should be the same for the MobileCoin blockchain no matter how many token types are supported. Namely, the sender, the recipient, and the amount should remain private to all except the participants in the transaction, and no double spending or unauthorized minting should be possible. In addition, we now specify that the asset type should remain private to all except the participants in the transaction.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A Transaction Output (TXO) has a confidential TokenType enum field (`masked_token_id`). Transactions with inputs and outputs, including a fee output, must be composed of a single TokenType. Some tokens may have additional functionality, such as minting and burning, which require verification of the TokenType as part of the authorization of the action.
```
eranrund, line 19: I don't believe its an enum, it's a bytes/Vec<u8> field.

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+    \\\ Token ID to be represented confidentially in each TXO
+    u32 token_id,
+    
+    \\\ Curve base point, unique for each TokenType, from which the TXO's commitment is 
+    \\\ generated with the value of the `amount` in the TXO
+    RistrettoPoint H_i
+}
+```
+
+The `masked_token_id` field is added to the [`Amount`](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/core/src/amount/mod.rs#L28) in each TXO, and functions similar to `masked_value`. The protocol representation of the TXO uses Protobuf, which enables backward compatability in this way: if the field is not populated (Protobuf bytes field of length 0), the default `token_id` is 0. Otherwise, the field is a Protobuf bytes field of length 4. Please see the [Upgrade Plan](#upgrade-plan) for more discussion on rolling out this change with consideration for backward compatibility.
```
eranrund, line 49: Nit: Might be worth linking to a specific commit, since as things evolve the could become stale.

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+transaction_3 | 111 | 3 | 2 | 6
+transaction_4 | 888 | 2 | 3 | 6
+
+Note how the two transactions paying more than their minimum fee are sorted to the top, in proportion to how much they are paying above the minimum fee, while all minimum fee transactions have an equal chance of beeing sorted because their `fee_ratio` is equal.
+
+### Performance Impact: Size & Space Analysis
+
+- Each TXO has 6 new bytes allocated in the Amount object for the `masked_token_id` (4 bytes for the value, 2 bytes for Protobuf overhead)
+- The RingCT bulletproof signature increases by at least 96 bytes to contain the Proof of Opening (plus Protobuf overhead).
+- Each block now contains a number of fee outputs scaling linearly with the number of transactions in a block, as opposed to a single fee output aggregated from multiple transactions in a block.
```
eranrund, line 199: This is only assuming we stop aggregating the fee outputs (which I would try to push against due to the bloating of the ledger). There was an alternative proposal - I unfortunately don't remember the specifics, but it was something along the lines of number of fee outputs = min(max token_id supported, number of transactions in block) (so there might be some 0 amount outputs).

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Confidentiality Analysis
+
+The [TxPrefix](https://github.com/mobilecoinfoundation/mobilecoin/blob/a9db0bc7bcac9bac1c1232b194003d02bccaf283/transaction/core/src/tx.rs#L149) contains the `token_id` and the `fee` in the clear. We, and any users of the protocol, need to clearly understand the footprint of this data, and the ways in which an attacker may try to discern the `token_id` of TXOs.
+
+The [TxPrefix](https://github.com/mobilecoinfoundation/mobilecoin/blob/edbc0162cbd647b6605fe817a8753508ca4515a2/transaction/core/src/tx.rs#L149) is included in the [Tx](https://github.com/mobilecoinfoundation/mobilecoin/blob/edbc0162cbd647b6605fe817a8753508ca4515a2/transaction/core/src/tx.rs#L103) that is delivered directly to the enclave through an attested channel at the [client\_tx_propose](https://github.com/mobilecoinfoundation/mobilecoin/blob/2d5190e60f6820cebcd585c19d16cecf9ba4d89b/consensus/enclave/impl/src/lib.rs#L299) endpoint. By attested, this means that the client would not deliver this payload unless the enclave's signed measurement matched the client's expectations.
```
eranrund, line 206:
```suggestion
The [TxPrefix](https://github.com/mobilecoinfoundation/mobilecoin/blob/edbc0162cbd647b6605fe817a8753508ca4515a2/transaction/core/src/tx.rs#L149) is included in the [Tx](https://github.com/mobilecoinfoundation/mobilecoin/blob/edbc0162cbd647b6605fe817a8753508ca4515a2/transaction/core/src/tx.rs#L103) that is delivered directly to the enclave through an attested channel at the [client_tx_propose](https://github.com/mobilecoinfoundation/mobilecoin/blob/2d5190e60f6820cebcd585c19d16cecf9ba4d89b/consensus/enclave/impl/src/lib.rs#L299) endpoint. By attested, this means that the client would not deliver this payload unless the enclave's signed measurement matched the client's expectations.
```

eranrund (commented):
I went over this and provided some nit comments. I am not fluent enough in the cryptography part to properly review that, but everything else seemed correct to me and is aligned with my understanding of @garbageslam's POC implementation and prior design discussions. 
Thank you for writing this!

UkoeHB:
I discussed this with Aaron Feickert, who wrote [Spats](https://github.com/AaronFeickert/spats) (WIP draft of confidential token ids for the Spark protocol).

Based on his work, I have a vision of a similar approach that permits hidden assets in the absence of fees (fees are a big problem for all hidden asset proposals, including this one, Spats, and this MCIP's original approach). Here is a brief sketch.

1. Define generator `H_ID = H_p("confidential_token_id")`.
2. Amount commitments are now: `C = x G + a H + token_id H_ID`. The generator is the same for all token types.
3. CLSAG proofs are unchanged.
4. Before doing range proofs, subtract `D = e G + token_id H_ID` from all output amount commitments. Range proofing the output commitments will show that removing `D` fully removed the `token_id` from those commitments.
5. Before doing the balance proof (`sum(pseudo commitments) - sum(output commitments) == 0`), subtract `D` from all pseudo commitments. The balance proof in combination with range proofing output commitments, and the assumption that `sum(token_id)` will never overflow (can be easily done by constraining new asset types to `< 2^64`), will mean that all `pseudo commitments - D` have no `token_id` component. In other words, if the range proofs and balance proof succeed, then all inputs/outputs have the same `token_id`.
6. To complete the balance proof, we must prove that `D` is well formed. Do a composition proof showing that `D = e G + token_id H_ID` (e.g. a Chaum-style proof as in Spats, or a composition-style proof as in Seraphis). This should be very cheap, only around 64 additional bytes I think. This composition proof is only needed if you want to hide `token_id` from verifiers. Mobilecoin may want to expose `token_id` to the SGX machine, in which case only `token_id` needs to be stored in the tx (set `e = 0` and don't bother with recording `D` since it can be computed as `D = token_id H_ID`).

isis-mc (commented):
Looks mostly good, just a couple inline comments and suggestions.

text/0000-confidential-token-ids.md
garbageslam: > This would seem to imply that, as additional new token types are added, that each token type's basepoint must be tried separately to recover the opening to the commitment. Is that accurate or did I misunderstand something?

No, that's not the proposal. What you do is, you unmask the `masked_token_id`, which is based on the assumption that this `TxOut` belongs to you, and so your shared secret computation is correct, so that you can compute the token id mask. Then you get a token id which is the only one it can be if it belongs to you, so you only have to test one value.

text/0000-confidential-token-ids.md
mfaulk: Good observation! I think you're right that the proof-of-opening is redundant.

Part of me still likes the proof-of-opening, because it directly asserts the transaction requirement that each input value is of the form _G + _ H_i. In my mind, the range proofs exist to prevent numerical overflow. Maybe it's fine if they are also used to assert that all inputs and outputs are of a single token type, but I worry that that kind of implicit claim is the kind of thing that tends to get overlooked if, say, we change the way we implement range proofs, or if we ever decide to use "vector Pedersen commitments" with mixed types (i.e. _G + _ H_i + _ H_k).

I don't feel very strongly about this; just thinking out loud

text/0000-confidential-token-ids.md
garbageslam: > This might in practise be fine, but I do not think it meets the standard cryptographic definition of orthogonal, since it depends at each stage upon the original Ristretto basepoint in compressed form.

I guess I don't agree: what it depends on is the assumption that `i -> Blake2b( RistrettoBasePoint XOR i)` is hard to distinguish from a random function in some sense. I would be happy with this assumption for any particular bytes in place of `RistrettoBasePoint`, just as I am happy with any particular domain separator string when hashing. Those bytes are only there because one of the requirements is that the value of this function for `i = 0` matches what we are using for MOB right now.

It seems like `i -> Blake2b( RistrettoBasePoint XOR i)` is just as good of a "nothing up my sleeve sequence" as digits of pi, they are both random arbitrary strings that are hard to control. If we went with digits of pi then this would no longer be efficient to compute for large values of `i`, and it needs to be efficient to compute for large values of `i`.

> (Also the Ristretto flavour of the Elligator2 mapping is not meant to be publicly exposed; is this something that was changed in a fork?)

I misread the docu for `RistrettoPoint::from_hash`, I guess it's actually taking a 64 byte hash and then doing what you said.

https://docs.rs/curve25519-dalek/latest/src/curve25519_dalek/ristretto.rs.html#704-713

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+To derive a sorting order without revealing the transaction type, the `fee` in the WellFormedTxContext becomes a `fee_ratio`, which is calculated based on the ratio between the minimum fees configured by the node operator on startup. For example, the network operators may have configured the following fees:
```
UkoeHB, line 152: Tbh I don't think you should allow fees outside the base asset type (MOB). It's really asking for trouble - how can you allow user-defined asset types? What if an asset type is completely worthless? Fees exist to mitigate DDOS, and must be carefully regulated by node operators for that purpose. The only robust way to do that is to coordinate on a single fee asset type.

Imo it would be better to require all special-asset-transfer txs have two sections: a 'base asset transfer' section, and a 'special asset transfer' section. The advantage here is you don't have to reveal the special asset type to validators. This can be done pretty trivially, since with my proposal all asset types can be CLSAGd and range proofed together (you only need to segregate inputs/outputs a little bit in order to do balance proofs).

garbageslam:
@UkoeHB thanks for discussing with Aaron and posting your thoughts (https://github.com/mobilecoinfoundation/mcips/pull/25#issuecomment-1033194827)

I'm trying to assess if this will be more or fewer elliptic curve operations if we do it as you suggest. I think several of us are going to think about your version and write back again soon.

text/0000-confidential-token-ids.md
garbageslam: I think it could work, because they could learn the priority they need to shoot for, and then relate that to the minimum fee for the token id they are trying to transact in?

UkoeHB: Users would be shooting in the dark... what happens if someone is spamming high fee txs? Users would have no way to accurately estimate the fee they need to get an urgent tx into the chain. This gives me the willies just thinking about it.. a fee market without a market.

garbageslam: > How can users decide what fee to use if fees are hidden (that seems to be your goal here)?

The goal is not that fees are hidden, but we do want the token id to be hidden. To the extent that different token ids have different minimum fees, the fee could reveal the token id. That's what we want to avoid.

> Users would be shooting in the dark... what happens if someone is spamming high fee txs? Users would have no way to accurately estimate the fee they need to get an urgent tx into the chain. This gives me the willies just thinking about it.. a fee market without a market.

The "priority" of the transaction (computed as fee / minimum fee) is not a secret, and the recently accepted priority values are published to the users to guide their fee selection. That's how the users can determine what fee to use to submit an urgent tx.

garbageslam:
This file should be `0025-...` not `0000-...`

text/0000-confidential-token-ids.md
sugargoat: Incorporating the text from @garbageslam's initial suggestion - should I incorporate any changes based on the discussion in this point, @isis-mc?

sugargoat: Incorporating the text from @garbageslam's initial suggestion - should I incorporate any changes based on the discussion in this point, @isis-mc?

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+    RistrettoPoint H_i
+}
+```
+
+The `masked_token_id` field is added to the [`Amount`](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/core/src/amount/mod.rs#L28) in each TXO, and functions similar to `masked_value`. The protocol representation of the TXO uses Protobuf, which enables backward compatability in this way: if the field is not populated (Protobuf bytes field of length 0), the default `token_id` is 0. Otherwise, the field is a Protobuf bytes field of length 4. Please see the [Upgrade Plan](#upgrade-plan) for more discussion on rolling out this change with consideration for backward compatibility.
+
+- To compute the 4 byte `masked_token_id` from the `token_id`, we hash the TXO shared secret, with a prefix, XORed with the `token_id`, and take the little endian representation of those bytes.
+
+    ```
+    masked_token_id = (token_id ^ Blake2B(token_id_tag | shared_secret)).to_le()
```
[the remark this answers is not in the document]
sugargoat: `token_id_tag` is meant to be a domain tag - some string, and | was meant to denote concatenation - definitely open to sugs for fixing up notation for clarity!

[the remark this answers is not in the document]
sugargoat: `token_id_tag` is meant to be a domain tag - some string, and | was meant to denote concatenation - definitely open to sugs for fixing up notation for clarity!

text/0000-confidential-token-ids.md
sugargoat: The backslash was necessary for rendering in my markdown editor - I'll verify in Github's rendering

text/0000-confidential-token-ids.md
sugargoat: You were correct! And we also removed the Proof of Opening from the proposal

text/0000-confidential-token-ids.md
sugargoat: We decided to remove the proof of opening, so marking this conversation resolved!

text/0000-confidential-token-ids.md
sugargoat: @UkoeHB - added a note at the bottom of the fee section to address how users may determine the appropriate fee given network congestion. Let us know what you think!

text/0000-confidential-token-ids.md
sugargoat: Updated to include @UkoeHB's description in the *Pederesen Generators* section! (It may be too verbose now - very happy to accept edits!)

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+Each TokenType has an explicit minimum fee specified by node operators, and included in the attestation handshake, ensuring that all nodes in the network are configured with the same fee minimums and TokenType sets. New TokenTypes can be added only with unanimous agreement from all node operators. 
+
+A transaction specifies the TokenType for the entire transaction, and includes a "Proof of Opening" that all `token_ids` in the inputs and outputs are the same. This prevents spending one TokenType as a member of another TokenType. The TransactionBuilder now takes a `token_id` as an argument in its initializer. The TokenType is set in the clear on the TxPrefix for a transaction, which is only accessed in the enclave, after the payload is delivered via an encrypted, attested channel. This is similar to the TxInputs, and thus has a similar threat model for confidentiality. See [Confidentiality Analysis](#confidentiality-analysis) for an assessment of risk.
```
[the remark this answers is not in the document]
sugargoat: Thank you! We ended up removing the proof of opening, so resolving!

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+    - Note also, the blinding factor base point does not need to be unique per TokenType, so we use the same blinding factor base point for all TokenTypes.
+    - Note finally, the Proof of Opening requires all commitments in a transaction were generated using the same group generator, base point `H_i`. If the commitments in input and/or output TXOs in a transaction are constructed with different group generators, and the sender is able to unwrap the sum as specified in the protocol below, it implies that the sender knows a nontrivial linear relationship among the `H_i`, which contradicts the assumption that the `H_i` are orthogonal. 
+
+The non-interactive zero knowledge Proof of Opening protocol is the following: (See [Zero Knowledge Proofs and Commitment Schemes, Page 27](https://www.cs.purdue.edu/homes/ninghui/courses/555_Spring12/handouts/555_Spring12_topic23.pdf))
+
+
+1. The prover establishes a value, `d`, calculated by hashing their secret values from the commitment (the `token_id` and the `blinding_factor`). Implementation note, we may use [Merlin](https://docs.rs/merlin/latest/merlin/) rather than Blake2B.
```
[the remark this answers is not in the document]
sugargoat: We ultimately did not use the proof of opening approach, but we are using Blake2b for hashing, which matches how we implemented the `masked_amount`. I would like to better understand the tradeoffs for that use case and why we chose Blake2b, if @garbageslam or @isis-mc have thoughts?

text/0000-confidential-token-ids.md
```diff
@@ -0,0 +1,270 @@
...
+
+    ```
+    y = Blake2B(proof_tag_t | token_id)
+    s = Blake2B(proof_tag_b | blinding_factor)
+    d = G * y + H * s
+    ```
+
+2. The non-interactive verifier chooses challenge, `e`, by hashing the commitment and the value `d` from the previous step.
```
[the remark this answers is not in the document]
sugargoat: We ended up not going with proof of opening, so resolving!

text/0000-confidential-token-ids.md
sugargoat: This, and the proposal above, have nice properties. I may want to grab some time with you to flesh out my understanding a bit more!

text/0000-confidential-token-ids.md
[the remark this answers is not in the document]
garbageslam: the security requirement here is that the masked value must be indistinguishable from uniformly random bytes from the point of view of someone who doesn't know the shared secret.

If we assume blake2b has a "secret prefix prf" property, then I think it does that at > 128 bit security level. And that assumption is reasonable and already used elsewhere.

Merlin is another tool that we can use to do this.

I coded it using blake2b because we are already doing the value mask that way, so it seemed more consistent with the existing code.

text/0000-confidential-token-ids.md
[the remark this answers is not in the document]
garbageslam: Maybe instead of `token_id_tag` just call it `domain_tag` and the reader will understand?

I think `||` is more commonly used to denote concatenation in these kinds of docus, that would be my sug

text/0000-confidential-token-ids.md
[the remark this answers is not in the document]
sugargoat: Thank you! Will incorporate in the MCIP explanation in alternatives

text/0000-confidential-token-ids.md
UkoeHB: sure

garbageslam:
@UkoeHB I think a couple people are trying to write a post back about what advantages we see in having each token be its own component in the pedersen commitment, but one reason is we can build on it in a way described in MCIP #31 

* It makes it possible to have transactions with multiple token types in a relatively simple way
* With a few more small changes, that enables us to do atomic swaps on chain in a relatively straightforward way

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,220 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A Transaction Output (TxOut) has a confidential `token_id` bytes field (`masked_token_id`). Transactions with inputs and outputs, including a fee output, must be composed of a single `token_id`. Some tokens may have additional functionality, such as minting and burning, which require verification of the `token_id` as part of the authorization of the action.
+
+Each `token_id` has an explicit minimum fee specified by node operators, and included in the attestation handshake, ensuring that all nodes in the network are configured with the same fee minimums and `token_id` sets. New `token_ids` can be added only with unanimous agreement from all node operators.
```
UkoeHB, line 21: IMO centralized token issuance would hamstring MobileCoin as a 'multi-asset ecosystem'. A big part of ethereum's success derives from users having free reign to [define their own tokens](https://steemit.com/ethereum/@maxnachamkin/how-to-create-your-own-ethereum-token-in-an-hour-erc20-verified). Not to mention the added burden on node operators to validate/justify hard forking for every new asset.

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,220 @@
...
+
+A transaction specifies the `token_id` for the entire transaction, with a cryptographic guarantee that all inputs and outputs are of the same `token_id`. This prevents spending one `token_id` as a member of another `token_id`. The TransactionBuilder now takes a `token_id` as an argument in its initializer. The `token_id` is set in the clear on the `TxPrefix` for a transaction, which is only accessed in the enclave, after the payload is delivered via an encrypted, attested channel. This is similar to the `TxInputs`, and thus has a similar threat model for confidentiality. See [Confidentiality Analysis](#confidentiality-analysis) for an assessment of risk.
```
UkoeHB, line 23: If tokens are not truly confidential from the beginning, then there will likely be no way to remove SGX without also removing the 'confidential' from 'confidential tokens'.

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,220 @@
...
+
+For example, an implementation to obtain a generator for `token_id = i` could look like:
+
+```
+B = RistrettoPoint::from_hash(Blake2b(HASH_TO_POINT_DOMAIN_TAG | RISTRETTO_BASEPOINT ^ token_id))
+B_blinding = RISTRETTO_BASEPOINT
+```
+
+After this change, `range_proofs`, `rct_bulletproofs`, and `ring_signature/mlsags` will be relative to a Pedersen generator, and it is not possible to construct a range proof relative to another generator if those generators are orthogonal. Thus, it is guaranteed that all transaction inputs and outputs are using the same `token_ids`. This is implied by the homomorphic encryption property of [Pedersen Commitments](https://www.cs.cornell.edu/courses/cs754/2001fa/129.PDF), namely that addition on the commitments preserves the additive relationship of the pre-committed values, as long as they are using the same group generator. `H_i` is revealed to the verifier, who must show that the sum of inputs does not overflow, or `sum(outputs) = a * G + b * H_i` for some TxOut-specific values of `a` and `b`.
```
UkoeHB, line 90: Modifying range proof structures to accommodate this proposal is intrusive, prevents efficient batch-verification which directly affects the upper bound of tx throughput, and may impose a burden on/barrier to future proof upgrades.

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,220 @@
...
+    - Note also, the blinding factor base point does not need to be unique per `token_id`, so we use the same blinding factor base point for all `token_ids`.
+
+### Fee Outputs
+
+Currently, the enclave aggregates the fees for multiple transactions in a block, in order to mint a single fee output for the block and conserve space. To support multiple confidential `token_ids`, when minting fee outputs, the enclave must create multiple fee outputs if multiple `token_ids` were used in the block. A consideration here is that we would like for an observer not to discern whether a block contains transactions of multiple asset types by statistically analyzing the number of outputs to derive information about the number of fee outputs. 
+
+A simple proposal is that each block now contains a number of fee outputs scaling linearly with the total number of supported `token_ids`. This way we can continue to aggregate fees, but not reveal how many token types were included in the block. For blocks with fewer transactions than `token_ids`, the number of fee outputs is `min(num_token_ids, num_transactions_in_block)`
```
UkoeHB, line 108: This greatly increases fee output spam (which has privacy implications), and prevents [fee epochs](https://github.com/mobilecoinfoundation/mcips/pull/10) as mentioned previously.

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,220 @@
...
+
+To derive a sorting order without revealing the transaction type, the `fee` in the `WellFormedTxContext` derives a `priority`, which is calculated based on some ratio between the minimum fees configured by the node operator on startup. For example, the network operators may have configured the following minimum fees:
```
UkoeHB, line 116: This may allow node operators to set up a token-id oracle.

UkoeHB (commented):
Second review

text/0025-confidential-token-ids.md
garbageslam: I think this is correct -- we want users to be able to define their own tokens. But as you have pointed out elsewhere, if those tokens can be used to pay transaction fees, then transaction fees become ineffective as a rate-limiting / DOS prevention mechanism.

However, we believe the following:
* In many cases, users of a payment rail would rather conduct transactions in something pegged to their local fiat currency
* Requiring them to purchase MOB in order to use our network is a negative user experience, and it adds another step to on-boarding. It also adds complexity because now the user has two balances instead of one.
* Therefore, we would rather let them pay the fee in their currency of choice, and the foundation will deal with the conversions on the backend. As long as they pay the foundation something that we know has value, then the fee serves its purpose of DOS mitigation.

So this together suggests that we want to have some way to create alternate tokens that can be used to pay network fees, but we don't want just anyone to be able to stand up such tokens.

So where I hope we will end up is, there will eventually be a way to allow people to create their own tokens, but not all tokens will be allowed to be used to pay fees, and only tokens approved by the foundation will be allowed to do that.

We haven't put much thought yet into what that part of it looks like. If you have suggestions, would love to hear.

text/0025-confidential-token-ids.md
garbageslam: Right now I think we believe it can be done using ZK-STARK's. I expect that if we accept this MCIP, then this would be a blocker for removing SGX.

text/0025-confidential-token-ids.md
garbageslam: That's a good point, and a reason to do it the way the Spats proposal works.

I think right now we don't have any of the batch verification stuff in place, so we'll have to think seriously about this. I don't know if this is actually the bottleneck right now in transaction validation.

text/0025-confidential-token-ids.md
garbageslam: 1. I guess I'm not sure that this is greatly increasing fee output spam. We currently only have one transaction per block typically. So every 3rd transaction output or so is fee right now. I don't think this proposal can ever lead to more than this, and also in the case of heavy load, if there are hundreds of transactions in a block but many fewer token ids, then this will be better than that.
2. Can you explain why this prevent fee epochs? It seems to me that fee epochs would still work just fine if we decided it was necessary, we will just accumulate fees from several different token types.

UkoeHB: 1. Ah my mistake, 'greatly' isn't accurate for this PR, although PR #31 could make it significant.
2. Fee epochs requires printing the fee amount in block headers, so it would eliminate the confidentiality of token ids. Maybe you could use pedersen commitments or something, but to prevent node operators from obtaining a token id oracle you'd need one commitment (and masked amount/id) per extant token id, which leads to a lot of blockchain bloat (possibly even more bloat than current fee spam causes, if the list of token ids became long enough).

garbageslam: I see -- I didn't realize it was storing the fee amount in the block headers.

Is that necessary for it to work? What if we just stored in fee output in (ephemeral) enclave memory and wrote out some number of TxOut's every 1000 blocks or so as you suggest. 

It might be annoying in that we might have to give the enclaves a way to sync this state, so if nodes go down and come back up they will have the right accumulated fee values. I would kind of like them to sync up anyways on the hash of the latest block, so that doesn't seem too onerous to also sync the accumulated fee values.

The worst case would be that all the enclaves went down and some accumulated fees were lost. But, that might be acceptable, we've tried to engineer it all so that never happens anyways.

garbageslam: Idk what the path would be here when SGX is removed though, that seems pretty tricky.

UkoeHB: > It might be annoying in that we might have to give the enclaves a way to sync this state, so if nodes go down and come back up they will have the right accumulated fee values. I would kind of like them to sync up anyways on the hash of the latest block, so that doesn't seem too onerous to also sync the accumulated fee values.

I feel like this would be an easy way to break the network, since it sounds like you don't want to use SCP for syncing... (which is designed to prevent breakage).

text/0025-confidential-token-ids.md
garbageslam: We're trying to avoid that by normalizing based on minimum fee, usually the users just always use the minimum fee, and then priority would reveal no information.

It is true that we would not have this problem if MOB were the only thing that could be used for fees.

UkoeHB: A node operator can set a tx's priority to an arbitrary value by configuring the minimum fees on startup. This lets them guess-and-check the sorting to find the tx's token id.

garbageslam: > A node operator can set a tx's priority to an arbitrary value by configuring the minimum fees on startup

That's true but their node can't consense with other nodes unless they have the same configuration as them, because the config is hashed and then appended to the responder id used for node-to-node attestation. This means the the nodes won't be able to pair to eachother unless they have the same config.

So in real life, the attacker has to hard fork the network, and control all the nodes that the user is submitting to, in order to conduct this guess-and-check attack.

Still, I agree it would be better if this weren't possible at all.

text/0025-confidential-token-ids.md
garbageslam: For a lot of configuration right now, such as minimum fee settings, the strategy is to use start-up parameters that become part of the responder id, and leverage node-to-node attestation to ensure that the nodes are all configured the same way. This is because it is simpler to do it this way right now, than to use SCP to consense on config. (We would like to get there though.)

I don't think this actually makes the network more brittle -- it makes problems surface immediately if nodes are misconfigured because there is no way a node can proceed if it can't attest to the other nodes.

This is not dynamic state, which is more complex for sure. But it is an example of using attestation to ensure that some data is being shared correctly.

The reason I'm not sure we need to "consense on accumulated fees" is, the accumulated fees are strictly a function of the transactions in each block. And we are already using SCP to consense on that. So we can just compute (in the enclave) based on what we already consensed on, and we don't need to do a second round of consensus to agree on the (deterministic) computation of fees.

The only part about this that is different from what we already do is, when a node restarts, its enclave loses all state. Currently, nodes that fall far enough behind go into a catchup mode where they trust blocks based on peer signatures and don't expect to see the Tx's for those blocks. Eventually they finish catching up and then they are validating transactions in the enclave again. However, if we tried to do fee epochs with secret accumulation in the enclaves, they would have to recover the accumulated fees as well from their peers, when they exit catchup. I don't think this would be a problem though if the accumulated fees are transmitted using node to node attestation.

text/0025-confidential-token-ids.md
UkoeHB: I see, then I think this part of the document should be clarified, since the way it's written it sounds like nodes can freely define their own startup values.

text/0025-confidential-token-ids.md
UkoeHB: IIRC, node catch-up (state recovery) relies on SCP (see mechanics of mobilecoin section 9.4.1).

garbageslam: it relies on SCP in that, the decision to trust blocks based on peer signatures is based on if those peers form a quorum.

i think it is also valid to send fee counters over node-to-node attested channels, based on a group of peers that form a quorum

UkoeHB: From the meeting: @garbageslam mentioned that you could use fee priorities to build a token ID oracle from user-transmitted transactions, since users don't validate node configuration.

UkoeHB:
Update on my previous [proposed scheme](https://github.com/mobilecoinfoundation/mcips/pull/25#issuecomment-1033194827): an observer would be able to divide the output set of each transaction into two groups, 'fee asset' outputs and 'non-fee asset' outputs. If fees are all in MOB, then the full ledger output set can be divided into 'base asset' outputs and 'secondary asset' outputs. This would reduce the efficacy of ring signatures unless users could use that distinction to only choose ring members that match their inputs' asset types (base vs secondary).

garbageslam:
@sugargoat can you please update this docu at some point to include the fixes to the extended message when signing ring MLSAG? https://github.com/mobilecoinfoundation/mobilecoin/pull/1700/files/c926556e63220c754b29a05b2c72fd5f39ef0e26#r836646544

garbageslam (changes requested):
Please also state the block version that this feature is targeted to release in

UkoeHB:
[This](https://eprint.iacr.org/2022/510) new paper (Bulletproofs++) mentions 'typed' range proofs for multi-asset protocols.

garbageslam:
I created a pull request here with updates and improvements:

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,339 @@
...
+
+# Summary
+[summary]: #summary
+
+Enable new token types to be added to the protocol and transacted on chain without revealing the token type.
+
+# Motivation
+[motivation]: #motivation
+
+The MobileCoin ledger protocol currently supports one token type, MOB. We would like to allow the addition of assets to the ledger without revealing which assets were involved in any transaction. The transaction confidentiality and integrity guarantees should be the same for the MobileCoin blockchain no matter how many token types are supported.
```
garbageslam, line 14:
```suggestion
The MobileCoin ledger protocol currently supports one token type, MOB. We would like to allow the addition of new assets to the ledger. The transaction confidentiality and integrity guarantees should be the same for the MobileCoin blockchain no matter how many token types are supported.
```

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,339 @@
...
+
+A `token_id` is a 64-bit number. Every `TxOut` will now have a corresponding `token_id`, and this will be committed to the chain via a new `masked_token_id` field.
+The `token_id` of a `TxOut` is recovered by a `TxOut` owner in a similar way to the `value`, during view key matching.
+
+The transaction object may now spend inputs and outputs of any token id, but in a single transaction all inputs and outputs must have the same `token_id`.
+
+The fee for such a transaction is also paid in that `token_id`. This is now specified in the `fee_token_id` field in the `TxPrefix`, and implies the token id for all inputs and outputs.
+
+The minimum fee varies from token id to token id. This is configured by network operators at startup, and hashed into the responder id during node-to-node attestation to ensure that the network
+has one uniform value for minimum fee.
```
garbageslam, line 36:
```suggestion
has one uniform value for minimum fee for each token.
```

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,339 @@
...
+to sort the transactions.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Recap
+
+Recall that in current revisions of MobileCoin, the value of a `TxOut` impacts the fields of the `TxOut` in two ways:
+
+* It appears in the masked value, where it bytes have been XOR'ed with the result of hashing the `TxOut` shared-secret
```
garbageslam, line 49:
```suggestion
* It appears in the `masked_value`. Here the bytes of the u64 value have been XOR'ed with the result of hashing the `TxOut` shared-secret
```

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,339 @@
...
+* It appears in the Pedersen commitment (`Amount::Commitment`), which has the form `v * H + v_blinding * G`
+
+Here `H` is a curve point derived by hashing to curve, and `G` is the ristretto basepoint. `v_blinding`, the blinding factor,
+is also derived by a hash of the `TxOut` shared secret.
+
+During view key matching, the recipient of a `TxOut` attempts to compute the shared secret, by key exchange with their view private key
+against the `TxOut` public key. They cannot confirm that this shared secret computation was successful -- however, if it is successful,
```
garbageslam, line 56:
```suggestion
against the `TxOut` public key. They cannot confirm directly that this shared secret computation was successful -- however, if it is successful,
```

garbageslam (approved):
LGTM, I had only minor suggestions

text/0025-confidential-token-ids.md
```diff
@@ -0,0 +1,339 @@
...
+
+The fee for such a transaction is also paid in that `token_id`. This is now specified in the `fee_token_id` field in the `TxPrefix`, and implies the token id for all inputs and outputs.
+
+The minimum fee varies from token id to token id. This is configured by network operators at startup, and hashed into the responder id during node-to-node attestation to ensure that the network
+has one uniform value for minimum fee.
+
+Previously, the fee of a `Tx` is exposed directly to the nodes (outside of the enclave) so that they can sort transactions by fee. After this change, the fee value may reveal information about the
+token id, because different tokens have different minimum fees. To mitigate this, the enclave divides the fee by the minimum fee, deriving a new value called `priority` which is revealed to the node
+to sort the transactions.
+
```
garbageslam, line 41:
```suggestion
For technical reasons, the minimum fee of every token now has additional requirements:

* It must be at least 128
* It must be evenly divisible by 128

```

eranrund (approved):

PR #26, opened 2022-02-06
Block-Version-based Protocol Evolution

garbageslam:
A proposal to use `block_version` to control schema evolution

[Rendered MCIP](https://github.com/garbageslam/mcips/blob/encrypted-memos/text/0024-encrypted-memos.md)


text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,228 @@
...
+
+To make this concrete, this means that if we want to roll out changes that add new
+features to `TxOut`'s, the new clients and old clients probably cannot coexist the way
+they might in a more conventional application.
+
+If a new client sends a `TxOut` to a user using an old client, which contains a `TxOut`
+feature that they don't recognize, then:
+1. They may fail to compute the correct block id
+2. If they submit a transaction spending the TxOut, the merkle proof of membership
+   may be wrong if the hash doesn't incorporate the new field.
```
UkoeHB, line 46: This is a good criticism of the dual-rule blocks approach, that I sadly did not foresee.

eranrund (approved):
LGTM.
I think that for coordinating block_version bumps, there are a few options that I can think of, all not terribly complicated to implement:
1) Command line configuration flag to the nodes that gets hashed into the responder id, similar to how we coordinate overriding the minimum fee
2) Adding a new transaction type to SCP - we have a POC for that (although in terms of code, this is the most complicated approach)
3) Deciding in advance on a date where the change happens, and have the nodes switch automatically at the given date/time. This is a bit janky since there is no way to ensure all nodes clocks are perfectly synced, so this will likely need to involve responder id change or something similar to that as well (to prevent nodes using the old block version from talking to nodes that have switched to the new block version).
Either way, I am confident that this part is a solvable problem.
Thanks for writing this MCIP!

text/0026-block-version-based-protocol-evolution.md
UkoeHB: @garbageslam What if membership proofs just hashed the core TxOut components? I.e. the onetime address and amount commitment. Then you can change memo fields arbitrarily without old clients needing to know about it when constructing a membership proof for the old enclave in a 'dual-rule' block. At the very least, you'd only need to do single-rule transitions on the rare occasion that core TxOut components change (if ever).

garbageslam: > This is a good criticism of the dual-rule blocks approach, that I sadly did not foresee.

Yeah I didn't really foresee this either. I saw pieces of it but not the whole thing. I think James saw the problem a while ago and it wasn't until now that we had a clear plan how to fix it.

garbageslam: > At the very least, you'd only need to do single-rule transitions on the rare occasion that core TxOut components change (if ever).

I think the simplest way to make this work without this kind of "dual rule" stuff would have been if:

* Instead of mc-crypto-digestible, we hash raw protobuf payloads, and clients leave all the fields in there for this hash even if they don't recognize some of them.
* Because protobuf serializations are not canonical, we would have to either make a "canonical protobuf" serializer, or, a routine that takes protobufs and makes them canonical.

It's a little late in the game for that. Making a protobuf serializer is really complicated, and we didn't really have the resources to try to do that, esp. that would support multiple languages, when we had only a few engineers. The `mc-crypto-digestible` thing was a lot easier to put together. We could also have tried to make a routine that takes serialized protobuf payloads and reorders the fields to make them canonical, but that also would be pretty complicated.

UkoeHB: > I think it still may break clients that are syncing the whole ledger, because if they are on an old block version, and TxOut's appear that have new fields, they may compute the wrong block id and then reject the block. So that would instantly force those older clients to update or be broken. So I don't think that's a complete solution unfortunately, otherwise yeah it would have been my preference to do it that way

Yeah anyone syncing the ledger needs a full and updated codebase. I think it would be wasted effort to try to make universal backward compatibility across forks.

The goals of dual-rule blocks are: 1) prevent txs from failing if they are submitted right before a transition point, 2) let 'simple' codebases like wallets only worry about one enclave at a time with loose upgrade requirements around forks (i.e. wallets that aren't syncing the entire chain).

text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,228 @@
...
+  ```
+  and using these in the transaction validation and construction routines, which will get us most of the benefits
+  of a feature-flag based system from a programming point of view.
+
+# Prior art
+[prior-art]: #prior-art
+
+This proposal relates to several earlier MCIP proposals:
+
+* 0007-Protocol-version-statement-type
```
garbageslam, line 179: #7

text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,228 @@
...
+* 0008-SCP-based-hard-forks
```
garbageslam, line 180: #8

text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,228 @@
...
+* 0009-Dual-rule-blocks
```
garbageslam, line 181: #9

text/0026-block-version-based-protocol-evolution.md
garbageslam: > Yeah anyone syncing the ledger needs a full and updated codebase. I think it would be wasted effort to try to make universal backward compatibility of old clients across forks.

I think the model we want to have is like,
* Users are expected to update every X days. If you have a phone app, it will prompt you to approve updates periodically. If you are running a point-of-sale terminal, you may have to do scheduled maintenance. This is all normal.
* Users expect that as long as they do the upgrades on schedule, the system always works.

Suppose that e.g. a grocery store decided to do a trial run of MOB point-of-sale terminals, which are syncing the ledger probably, using full-service or mobilecoind let's say.

If the terminal is old and the customer has new versions of our software, and the terminal breaks (because it can't spend a TxOut with a new field), that's not a good experience for the merchant. Perhaps they can fix it by doing a update but it's not the experience they have with VISA for instance. However, if they update aggressively (and before their customers), then their customers are the ones having the unspendable TxOut's, if they have to process a refund or something. So it's a bit of a catch 22, someone is going to have a broken transaction and have to suddenly do an update faster than expected.

I think we'd really like to be able to ship clients that are guaranteed to work for their life-time and not break them prematurely, no matter if they are desktop wallets or mobile wallets. I don't personally think this is wasted effort, as long as we figure out a system that works and follow it, then everyone will have a good user experience.

UkoeHB: > I think we'd really like to be able to ship clients that are guaranteed to work for their life-time and not break them prematurely, no matter if they are desktop wallets or mobile wallets.

This can only be achieved for scanning the ledger for owned outputs (bare-minimum balance-checking where the only things you look at are onetime address, amount commitments, encoded amounts, output pubkeys, and key images), assuming you don't transition to new crypto for those pieces (e.g. Seraphis). Spending to a new enclave always requires an updated client, since you need to sign all tx data which means knowing all the current formats/etc. Reading miscellaneous new memos also always requires an updated client (obviously).

garbageslam: Yes but if we have a policy like, we distribute updates every X days that trust both current MRENCLAVE and next MRENCLAVE, and only after `X` days do we deploy the new enclaves, we can also navigate that without ever breaking clients that are less than `X` days old.

jcape:
> LGTM. I think that for coordinating block_version bumps, there are a few options that I can think of, all not terribly complicated to implement:
> 
> 1. Command line configuration flag to the nodes that gets hashed into the responder id, similar to how we coordinate overriding the minimum fee

I was kind of leaning towards this, as it seems like "do a coordinated restart" w/o enclave changes is the easiest way to make this go. That being said, I don't *think* there's a danger of "some nodes externalize v1, some nodes externalize v2", as they wouldn't be able to reach consensus on it until there's K/N agreement on what version things are writing.

One caveat is there will be a tombstone block age-out for v2 transactions submitted to v2 nodes before there's a quorum of v2 nodes (and similarly for v1 transactions to v1 nodes before the restart is complete).

One interesting aside to this is that we'll need to plumb the block version into something like `GetLastBlockInfo` for non-fog clients anyways, so while adding it to fog would be nice, it may be unnecessary.

garbageslam:
So like, one benefit of being able to programmatically trigger this would be, we could modify CD so that e.g. it deploys a network at block version 1, then run test client, then if it passes, trigger upgrade to block version 2, then run test client again and see if it passes. That would be kind of nice for testing that the upgrades don't break things pretty early in the development pipeline

samdealy (approved):
LGTM. Very well written and clear to me!

mfaulk (approved):
This is very well thought out!

text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,288 @@
...
+
+If a new client sends a `TxOut` to a user using an old client, which contains a `TxOut`
+feature that they don't recognize, then:
+1. They may fail to compute the correct block id
+2. If they submit a transaction spending the TxOut, the merkle proof of membership
+   may be wrong if the hash doesn't incorporate the new field.
+3. Even if they are a fog client, and (new) fog is building the proof of membership for them,
+   if the client strips out the fields it doesn't understand that fog gives them, then consensus
+   will likely reject the transaction. (There are some protobuf implementations that support "preserving"
+   unknown fields, but PROST, which is the only rust implementation which is enclave-compatible, does not do this.)
```
remoun, line 50: I wonder if it makes sense for us to invest in making Prost support preserving unknown fields.

text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,288 @@
...
+This proposal relates to several earlier MCIP proposals:
+
+* [0007-Protocol-version-statement-type](https://github.com/mobilecoinfoundation/mcips/pull/0007)
+* [0008-SCP-based-hard-forks](https://github.com/mobilecoinfoundation/mcips/pull/0008)
+* [0009-Dual-rule-blocks](https://github.com/mobilecoinfoundation/mcips/pull/0009)
+
+In particular, `0007-protocol-version-statement-type` proposes a way that node operators
+could trigger `block_version` to increase by voting for this and reaching consensus on it using SCP itself.
+
+This proposal would resolve some of the drawbacks in this proposal around ad-hoc mechanisms, and may strengthen this proposal.
```
remoun, line 247: Nit: "this proposal" twice in the same sentence.
```suggestion
That proposal would resolve some of the drawbacks in this proposal around ad-hoc mechanisms, and may strengthen this proposal.
```

remoun (approved):
LGTM

text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,288 @@
...
+Desktop clients, which use the lmdb database `mc-ledger-db`, already work by comparing the block version of new blocks to the
+`mc_transaction_core::BLOCK_VERSION` constant. They should continue doing this, but the constant should be renamed `MAX_BLOCK_VERSION`.
+
+We do not at this time propose that the `Tx` object itself should contain a `block_version`.
+Instead, transactions are valid if they conform to the rules for the current block version.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This adds complexity to the transaction validation code, which now must support multiple versions of the rules.
```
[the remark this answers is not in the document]
garbageslam: idk, I guess my thinking is:
* In the "initial" plans for how to do versioning, we didn't plan on doing this, there would just be one version for transaction core lib.
* We also planned to roll out the memos that way
* I'm arguing that we have to accept this complexity to avoid breaking the clients sometimes in this MCIP but I'm trying to treat that as a position of the MCIP rather than state it as a fact, and the reader can draw their conclusions

text/0026-block-version-based-protocol-evolution.md
```diff
@@ -0,0 +1,288 @@
...
+should both be `block_version` aware and support multiple block versions when an upgrade occurs.
+
+* The `block_version` of the blockchain is the `block_version` of the latest block.
+* The block version of a block can never be less than the `block_version` of its parent.
+* We propose that the encrypted memo field from `0003-encrypted-memos` MCIP should be associated to a bump to `block_version = 2`.
+  That is, consensus will reject `Tx` that contain any `TxOut` that have memos until `block_version = 2`.
+* We propose that the `masked_token_id` field from the  `0023-confidential-token-ids` MCIP should be associated to a bump to `block_version = 3`.
+  That is, consensus will reject `Tx` that contain any `TxOut` that have this field until `block_version = 3`.
+* We propose that `block_version = 3` may be triggered by the acceptance of the first minting transaction to the blockchain.
+* We do not at this point have a concrete proposal for how to trigger `block_version = 2`.
```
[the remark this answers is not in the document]
remoun: I think so? I'm imagining an internal UI with a button that kicks off that SCP statement.

PR #30, opened 2022-03-14
Standardized sender signature

cbeck88:
A proposal for a standardized signature for the sender of a TxOut, who may otherwise be anonymous.

[Rendered Proposal](https://github.com/garbageslam/mcips/blob/standardized-sender-sig/text/0030-standardized-sender-sig.md)

### DIFF
--- a/text/0030-standardized-sender-sig.md
+++ b/text/0030-standardized-sender-sig.md
@@ -26,13 +26,15 @@ However, this imposes fairly strict size requirements on the data that can be se
 
 Outside of the memo field, we have a notion of "receipts" and "confirmation numbers" which a sender
 can use to identify themselves to a recipient. These are hashes of the TxOut shared secret, which
-are similarly deniable, and convince the recipient that whoever sent them this number must have access
-to secrets involved in the construction of the TxOut.
+convince the recipient that whoever sent them this number must have access to secrets involved in the
+construction of the TxOut. These receipts have a "deniability" property also shared by signal messages.
+This arises because they are based on the TxOut shared secret, which can be formed by the sender OR the recipient.
 
-However, these numbers are not useful for convincing a third-party who is not a party to the transaction,
-because only the recipient is able to validate these numbers. The recipient could give their private keys
-away to a third party, but this entails giving up their privacy. There is currently no form of receipt that
-can be validated by anyone, just from the blockchain, without such measures.
+However, these confirmation numbers are not useful for convincing a third-party who is not a party to the transaction
+that a particular person sent the transaction, because only the transaction recipient is able to validate these numbers.
+The recipient could give their private keys away to a third party, but this entails giving up their privacy. Even then,
+because of the deniability property, this still only convinces the third party that one of two people sent the Tx.
+There is currently no form of receipt that can be validated by anyone, just from the blockchain.
 
 This is desirable by some app project developers, who would like to store encrypted metadata per TxOut,
 but off-chain. For example, they might like to have a database of encrypted metadata, keyed on TxOut's sent
@@ -77,6 +79,17 @@ We propose to
 
 # Prior art
 [prior-art]: #prior-art
+
+To recap on the deniability property:
+
+* Deniable authentication means that, when the sender "signs" a message, the signature is made by a key exchange with the recipient.
+  * This means that both the sender AND the recipient are technically capable of producing the signature.
+  * The recipient who sees the signature, and knows they didn't produce the signature, concludes the sender must have sent it.
+  * However, the recipient cannot convince another person that they did not forge the signature.
+  * As a result, if the recipient shows the message to a third party (to embarrass the sender), the sender can deny having written
+    the message, and claim that the recipient made the whole thing up.
+* A [good intro](https://www.praetorian.com/blog/an-opinionated-series-on-why-signal-protocol-is-well-designed-deniability/)
+* A (technical) Signal blog post on the topic: https://signal.org/blog/simplifying-otr-deniability/
 
 This proposal touches on some earlier MCIPS:
 
--- a/text/0030-standardized-sender-sig.md
+++ b/text/0030-standardized-sender-sig.md
@@ -31,7 +31,7 @@ to secrets involved in the construction of the TxOut.
 
 However, these numbers are useless for convincing a third-party who is not a party to the transaction,
 because only the recipient is able to validate these numbers. There is currently no form of receipt that
-can be validated just from the blockchain.
+can be validated by anyone, just from the blockchain.
 
 This is desirable by some app project developers, who would like to store encrypted metadata per TxOut,
 but off-chain. For example, they might like to have a database of encrypted metadata, keyed on TxOut's sent
--- a/text/0030-standardized-sender-sig.md
+++ b/text/0030-standardized-sender-sig.md
@@ -27,8 +27,9 @@ However, this imposes fairly strict size requirements on the data that can be se
 Outside of the memo field, we have a notion of "receipts" and "confirmation numbers" which a sender
 can use to identify themselves to a recipient. These are hashes of the TxOut shared secret, which
 convince the recipient that whoever sent them this number must have access to secrets involved in the
-construction of the TxOut. These receipts have a "deniability" property also shared by signal messages.
+construction of the TxOut. These confirmation numbers have a "deniability" property also shared by signal messages.
 This arises because they are based on the TxOut shared secret, which can be formed by the sender OR the recipient.
+(For more discussion, see the prior art section.)
 
 However, these confirmation numbers are not useful for convincing a third-party who is not a party to the transaction
 that a particular person sent the transaction, because only the transaction recipient is able to validate these numbers.
@@ -47,6 +48,7 @@ We can simply use the Schnorrkel-based signature with Ristretto curve points tha
 
 This might also be useful because it makes the notion of receipts somewhat more robust -- arbitrary metadata
 can be signed by the sender, and this form of signature will not have the deniability property.
+This happens because it is a traditional digital signature, based on a key that only the sender has.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
--- a/text/0030-standardized-sender-sig.md
+++ b/text/0030-standardized-sender-sig.md
@@ -107,6 +107,7 @@ This proposal touches on some earlier MCIPS:
 * [0018-Janus-attack-mitigation](https://github.com/mobilecoinfoundation/mcips/pull/0018)
 
 Outside of MobileCoin, there are many blockchain projects that want to store chain-related data off-chain.
+This is often motivated by scaling considerations.
 
 Popular strategies include:
 * Embed hash of the stored data into the transaction
--- a/text/0030-standardized-sender-sig.md
+++ b/text/0030-standardized-sender-sig.md
@@ -1,0 +1,137 @@
+- Feature Name: Standardized Sender Signature
+- Start Date: 2022-03-14
+- MCIP PR: [mobilecoinfoundation/mcips#0030](https://github.com/mobilecoinfoundation/mcips/pull/0030)
+- MobileCoin Epic: None
+
+# Summary
+[summary]: #summary
+
+Enact MCIP #18 (Janus Mitigation), which sets the ephemeral key in the encrypted fog hint
+to be the tx private key times the Ristretto base point. The original motivation was to enact
+Janus attack mitigation.
+
+We find a secondary use-case to motivate this: it becomes possible for the signer to create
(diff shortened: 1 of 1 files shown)


text/0030-standardized-sender-sig.md
```diff
@@ -0,0 +1,137 @@
...
+We find a secondary use-case to motivate this: it becomes possible for the signer to create
+a Schnorr signature over any payload, which a verifier can validate against the keys appearing
+in a `TxOut`, without revealing the identity of the signer. We propose to both do #18 and standardize
+the construction and validation of this signature.
+
+# Motivation
+[motivation]: #motivation
+
+Often in a payments flow, it is necessary for the sender and recipient to exchange additional messages
+and metadata about the payment. Frequently, these messages need to be authenticated.
```
sugargoat, line 22:
```suggestion
and metadata about the payment, on or off-chain. Frequently, these messages need to be authenticated.
```

text/0030-standardized-sender-sig.md
```diff
@@ -0,0 +1,137 @@
...
+
+We have a mechanism for doing this in the memo field, which attaches fixed-size payloads to TxOut's.
+However, this imposes fairly strict size requirements on the data that can be sent.
+
+Outside of the memo field, we have a notion of "receipts" and "confirmation numbers" which a sender
+can use to identify themselves to a recipient. These are hashes of the TxOut shared secret, which
+are similarly deniable, and convince the recipient that whoever sent them this number must have access
+to secrets involved in the construction of the TxOut.
+
+However, these numbers are useless for convincing a third-party who is not a party to the transaction,
```
sugargoat, line 32:
```suggestion
However, these numbers are only capable of convincing a third-party who also has access to the recipients' private keys,
```

Or something like this - indicating that a 3rd party could be enlisted to do these confirmations, but only if they are provided with the correct material to do the validation.

text/0030-standardized-sender-sig.md
```diff
@@ -0,0 +1,137 @@
...
+but off-chain. For example, they might like to have a database of encrypted metadata, keyed on TxOut's sent
+by users of their app, and be able to ensure that only the true sender of the TxOut can post metadata associated
+to the TxOut, without learning anything about the sender, recipient, amount, etc. of the TxOut.
+
+After this proposal, a straightforward way to do this is that their app can sign the encrypted metadata using
+the tx private key, and their server can check this signature against the public key in the encrypted fog hint.
+We can simply use the Schnorrkel-based signature with Ristretto curve points that we already use in fog public addresses.
+
+This might also be useful because it makes the notion of receipts somewhat more robust -- arbitrary metadata
+can be signed by the sender, and this form of signature will not have the deniability property.
```
sugargoat, line 46: Can you say more about why it does not have the deniability property?

text/0030-standardized-sender-sig.md
```diff
@@ -0,0 +1,137 @@
...
+
+* Make internal changes to the transaction builder, following MCIP #18
+  * Instead of being a second independent random private key, the private key used
+    for `mc-crypto-box` encryption will be the same as the tx private key for that
+    `TxOut`.
+  * Also, implement code for a "Janus test" which enables clients to detect the
+    Janus attack, as described in MCIP #18.
+* Standardize a Schnorrkel-based signature using Ristretto curve points, made using the tx private key,
+  and validated using the ephemeral key in the fog hint. This entails:
+  * Make the transaction builder either optionally return the tx private key when `add_output` is called,
```
sugargoat, line 71: Slight preference for signing at the time `add_output` is called, so that we don't have to build infrastructure around storing the private key for later use. Not sure whether we need to add anything to the MCIP about those tradeoffs, except maybe to consider the challenges of properly storing private key material in infrastructure.

sugargoat: The other side of the tradeoff is that if you want to send encrypted data about a previous TXO later, you may end up encrypting data for old TXOs when sending new TXOs, which could get messy to tease apart in the clients, or may not be so bad if we have a reasonable schema for this.

text/0030-standardized-sender-sig.md
```diff
@@ -0,0 +1,137 @@
...
+to MCIP #18, we think we should do MCIP #18 also, since it is valuable to mitigate
+Janus attacks and that MCIP is well-studied. It costs us almost nothing to start
+building transactions in the way described there, and at some point, flip the switch
+on clients actually flagging Janus test failures as suspicious.
+
+We do not believe right now that there is a simpler way to create this signature,
+as the signer does not know the root of the `public_key` or `target_key` of `TxOut`'s
+relative to the ristretto basepoint. The signer does know the root relative to some of
+the public subaddress keys of the recipient, but such a signature cannot be verified
+by someone who does not know who the recipient.
```
sugargoat, line 122:
```suggestion
by someone who does not know who the recipient is.
```

text/0030-standardized-sender-sig.md
garbageslam: So what I had in mind was not that infrastructure stores the key, but that the app can create and submit a transaction, and then sign something perhaps a second or two later that goes to the app server.

It might be that the app needs to hit some endpoint to get some of the data that will be signed. If we allow that they can hold onto the ephemeral private key after the `TransactionBuilder` is gone then they might be able to pipeline their calls effectively. If not then they might be forced to serialize them.

It's just a thought experiment, I don't know if it would matter in a real app.

text/0030-standardized-sender-sig.md
```diff
@@ -31,7 +31,7 @@ to secrets involved in the construction of the TxOut.
 
 However, these numbers are useless for convincing a third-party who is not a party to the transaction,
 because only the recipient is able to validate these numbers. There is currently no form of receipt that
-can be validated just from the blockchain.
+can be validated by anyone, just from the blockchain.
```
sugargoat, line 34: Is this an observation about the current state of things, or an indication that this MCIP solves this problem? My understanding is that this MCIP doesn't propose to have on-chain receipts that can be validated by anyone, but that the off-chain store could be, is that correct?

text/0030-standardized-sender-sig.md
garbageslam: The receipts based on the shared secret, have the deniability property, because both the sender and recipient can build the shared secret, and so form the signature. So it is technically possible for the recipient to have forged the signature, and for the sender to deny having created the receipt.

With this signature, the sender is signing it with the tx private key, which is a key only they have access to, and the recipient can't fake the signature using the view private key. So this signature is not deniable, it must have come from the sender.

text/0030-standardized-sender-sig.md
garbageslam: I feel like this doesn't make the motivation more clear -- i feel like the statement makes sense even in a non-blockchain context, and adding "on or off-chain" doesn't help

text/0030-standardized-sender-sig.md
garbageslam: fair!

garbageslam: i feel like, it is a bit weird to phrase things this way, because the recipient is not ever expected to give away their private keys, and if they are, that's an exceptional situation

text/0030-standardized-sender-sig.md
sugargoat: WFM!

text/0030-standardized-sender-sig.md
sugargoat: Good point - and that's one of our motivating design constraints. The "numbers are useless for convincing a third-party" felt false in that if I really needed to convince the third party, I could provide them my private keys. Maybe rephrasing to say we don't want to put users in a situation where they need to provide their private keys to a third party is a design goal of MobileCoin?

text/0030-standardized-sender-sig.md
sugargoat: Awesome - it may be worth including some of the text in your answer to define what we mean be deniability property

text/0030-standardized-sender-sig.md
sugargoat: Ok cool - that makes sense to me

sugargoat (approved):
LGTM!

text/0030-standardized-sender-sig.md
garbageslam: thank you!

jcape:
I'd like to get #18 fixed up and merged first, this would simply this MCIP and resolve the major issue: you can't implement this without directly referencing an MCIP that was never accepted.

Alternatively, the contents of #18 can get merged into this one.

text/0030-standardized-sender-sig.md
```diff
@@ -0,0 +1,162 @@
...
+
+# Summary
+[summary]: #summary
+
+Enact MCIP #18 (Janus Mitigation), which sets the ephemeral key in the encrypted fog hint
+to be the tx private key times the Ristretto base point. The original motivation was to enact
+Janus attack mitigation.
+
+We find a secondary use-case to motivate this: it becomes possible for the signer to create
+a Schnorr signature over any payload, which a verifier can validate against the keys appearing
```
mfaulk, line 14: To make sure I'm following this: the public key that a sender would sign against is in the encrypted fog hint, so only the recipient's fog server can obtain this key and use it to verify messages signed with it?

text/0030-standardized-sender-sig.md
```diff
@@ -0,0 +1,162 @@
+- Feature Name: Standardized Sender Signature
+- Start Date: 2022-03-14
+- MCIP PR: [mobilecoinfoundation/mcips#0030](https://github.com/mobilecoinfoundation/mcips/pull/0030)
+- MobileCoin Epic: None
+
+# Summary
+[summary]: #summary
+
+Enact MCIP #18 (Janus Mitigation), which sets the ephemeral key in the encrypted fog hint
```
mfaulk, line 9: We haven't merged MCIP 18, and it doesn't look like anyone has reviewed it yet. I think that needs to be reviewed and merged before we build on it. For example, I'm having a hard time finding the "fog hint ephemeral key" in the source, so it's hard for me to evaluate what's going on here.

mfaulk:
I'm wondering if the simplest thing to do here is to add a 'sender_signing_key' field to TxOut, which is a unique public key that is unrelated to tx_out.target_key and tx_out.public_key, and is unrelated to the fog hint. I know that would increase the size of the TxOut, but does that size increase really matter? On the flipside, it becomes extremely simple to say "this is the key that a sender should sign related data with", and anyone with access to the ledger can validate such signatures.

text/0030-standardized-sender-sig.md
garbageslam: so, the way that cryptobox works is:

* I get your public key (a Ristretto curve point)
* I choose a random scalar (private fog key, `r_fog` in Koe's notation in Mechanics of Mobilecoin)
* I do key exchange against your public key and `r_fog`, and I put `r_fog * G` as the first 32 bytes of the e_fog_hint.
* I use the key resulting from key exchange to encrypt and mac my payload
* I append 2 versioning bytes.

So the public key that a sender would sign against is `r_fog`, and can be validated against `r_fog * G`, which is the first 32 bytes of fog hint.

Koe proposed that instead of having two independent random keys `r` and `r_fog`, we should just set `r_fog = r`, and this enables a Janus mitigation test, without really impacting fog since the `r_fog` number is a perfectly fine secret key anyways.

So this breaks some encapsulation that's separating the `e_fog_hint` field from the rest of the Tx math. But, it serves several useful purposes, and it's not that big a deal from the point of cryptobox / fog.

garbageslam:
> I'm wondering if the simplest thing to do here is to add a 'sender_signing_key' field to TxOut, which is a unique public key that is unrelated to tx_out.target_key and tx_out.public_key, and is unrelated to the fog hint. I know that would increase the size of the TxOut, but does that size increase really matter?

It matters in that, it is a ledger format change, and an enclave change, and so would require a long turn around time to ship.

text/0030-standardized-sender-sig.md
garbageslam: For sure, we should definitely review that MCIP and merge it separately if we want to accept this MCIP, IMO

NotGyro:
This feels somewhat out of my depth, so this is probably a stupid question, but would the bytes signed include the memo field? I could see that being very useful.

garbageslam:
> This feels somewhat out of my depth, so this is probably a stupid question, but would the bytes signed include the memo field? I could see that being very useful.

This signature is mainly about, taking bytes and binding them to "whoever sent that Txout" without revealing who it is.

The memo field is already set by that person and it's bound to the TxOut by the blockchain hash, so if I'm understanding right I think signing those bytes would be redundant.

But, part of the point of this is to allow signing "additional" memo-like bytes for use in other contexts that don't have to appear on the chain.

PR #31, opened 2022-03-21
Transactions with Contingent Inputs

garbageslam:
A proposal to enable trustless atomic swaps.

[Rendered Proposal](https://github.com/garbageslam/mcips/blob/transactions_with_contingent_inputs/text/0031-transactions-with-contingent-inputs.md)

### DIFF
--- a/text/0031-transactions-with-contingent-inputs.md
+++ b/text/0031-transactions-with-contingent-inputs.md
@@ -42,8 +42,10 @@ requiring each party to "commit" their funds first, then trigger the swap, and h
 if the counterparty renegs. These protocols may require several rounds of interaction, and require several on-chain transactions, incurring fees and
 increasing the total delay for the swap to be finished. (Because of this complexity, alternatives have been sought such as "wrapped Bitcoin" which is an
 ERC20 representation of locked bitcoin which can be used more easily on the Ethereum network.) Another comment here is that in these multistep protocols,
-if the counterparty stops participating, you can still recover your funds, but you still pay two transaction fees to recover them from the escrow, and you
+if the counterparty stops participating, you can still recover your funds, but you still pay a transaction fees to recover them from the escrow, and you
 didn't manage to complete your swap. You can try again with a different counterparty, but if this happens repeatedly you will lose money to fees.
+If there are a lot of bad actors in the network, then this could happen often, and it may be very difficult for the network to prevent this from happening
+when the underlying mechanism is a multistep protocol.
 
 This proposal is considerably simpler -- the rounds of interaction are a minimum, requiring only one side to send a transaction fragment to another.
 No one is required to "escrow" their funds into a smart contract or temporary address. A party who desires to swap can broadcast their transaction fragment
--- a/text/0031-transactions-with-contingent-inputs.md
+++ b/text/0031-transactions-with-contingent-inputs.md
@@ -216,12 +216,9 @@ and if there are additional outputs in the `Tx` besides those appearing in the r
 transaction to become malleable -- if the process of proposing the Tx to consensus, a man in the middle were able to intercept it, they could tamper with these parts of the transaction, in a way that
 isn't possible for transactions currently.
 
-Since transactions are submitted over attested channels to the consensus enclave, there is already quite a lot of protection against tampering with in-flight transactions.
-However, another simple way to mitigate this is that a client can ensure that at least one input to the `Tx` does not have contingent input rules, and so signs the whole `TxPrefix`.
-
-Another possibility we considered is that inputs and outputs in a `TxPrefix` can be ordered together, and the MLSAG for an input always signs over everything that appears before it, but not over what appears later.
-However, this has a drawback, which is that if two people separately generate `SignedContingentInput`'s, then both inputs cannot be used in the same transaction.
-This is also a breaking change, while the actual proposal above allows for backwards compatibility.
+It turns out that it is hard even to build a transaction with properly balanced blinding factors if at least one input has not been presigned.
+Therefore, this should be required by all of the clients, and to completely avoid the malleability issues, the consensus network should verify that at least one input in the Tx does not have
+pre-signed rules.
 
 Finally, we describe the `SignedContingentInput` object.
 
--- a/text/0031-transactions-with-contingent-inputs.md
+++ b/text/0031-transactions-with-contingent-inputs.md
@@ -91,10 +91,11 @@ A "signed contingent input" is a payload consisting of:
 * A `RingMLSAG` signing this.
 * A token id, value, and blinding factor for the `pseudo_output_commitment`.
 * A token id, value, and blinding factor for any output in the `ContingentInputRules`.
+* A block version (which was in effect for the `RingMLSAG` signature).
 
 Anyone who owns a `TxOut` can use it to create a `SignedContingentInput`. The signed contingent input can be thought of as an offer
 to trade "this for that" -- you are willing to give away your `TxOut` if someone produces for you the `TxOut` that you specify.
-Once you sign the contingent input, you can give it away, and anyone who has it can fulfill your request to trade.
+Once you sign the contingent input, you can give that payload away, and anyone who has it can fulfill your request to trade.
 
 Giving away the signed contingent input does not reveal your public address, nor which `TxOut` in the ledger you actually own.
 It does reveal the token id and value of your pseudo-output, but this is unavoidable in any such scheme.
@@ -114,7 +115,8 @@ The `SignatureRctBulletProofs` object has two new fields, indicating token ids o
 message SignatureRctBulletproofs {
     repeated RingMLSAG ring_signatures = 1;
     repeated CompressedRistretto pseudo_output_commitments = 2;
-    bytes range_proofs = 3;
+    bytes range_proof = 3;
+    repeated bytes range_proofs = 4;
     repeated fixed32 pseudo_output_commitment_token_ids = 4;
     repeated fixed32 output_token_ids = 5;
 }
@@ -123,6 +125,11 @@ message SignatureRctBulletproofs {
 The `pseudo_output_commitment_token_ids` list must be the same length as `pseudo_output_commitments`,
 and `output_token_ids` must be the same length as `outputs` in the `TxPrefix`, or the transaction is invalid.
 (Alternatively, for backwards compatibility we could allow that if these lists are empty then they are filled with zeros.)
+
+We also need to add a `range_proofs` field to accommodate mixed transactions. This is because the current implementation
+of `bulletproofs` which we get from Dalek crate only allows one generator to be used for range proof aggregation. However,
+we believe that upstream could add support for using multiple generators in a single range proof, and then this complexity
+could be removed.
 
 In the `TxPrefix`, the `token_id` field is renamed to `fee_token_id`, and it only represents the token id of
 the fee output now.
@@ -152,10 +159,10 @@ and `output_token_ids`, rather than the `token_id` in the `TxPrefix`.
 
 To support contingent inputs, we introduce the following objects:
 
-A `ContingentInputRules` object is a protobuf with the following schema:
+An `InputRules` object is a protobuf with the following schema:
 
 ```
-message ContingentInputRules {
+message InputRules {
     // Outputs that must appear in the transaction for a transaction using this input to be valid
     repeated external.TxOut required_outputs = 1;
     // A maximum tombstone block value for the transaction. If zero, it means no maximum is enforced.
@@ -180,7 +187,7 @@ message TxIn {
     repeated TxOutMembershipProof proofs = 2;
     
     // Any rules which must be satisfied by the Tx that contains this TxIn
-    ContingentInputRules contingent_input_rules = 3;
+    InputRules contingent_input_rules = 3;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+- Tracking Issue: None
+
+# Summary
+[summary]: #summary
+
+After MCIP [#25](https://github.com/mobilecoinfoundation/mcips/pull/0025),
+evolve the transaction format with two new features:
+
+* Multi-asset transfers: A single Tx can now contain inputs and outputs of multiple different types. Each input and output must be annotated with its type,
+to enable range proof validation. As long as the Tx is balanced across all types, it is valid. The transaction fee can be paid in with any of the token types that are used.
```
sugargoat, line 13:
```suggestion
to enable range proof validation. As long as the Tx is balanced across all types, it is valid. The transaction fee can be paid with any of the token types that are used.
```

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+# Motivation
+[motivation]: #motivation
+
+Contingent inputs are interesting because they allow users to construct "transaction fragments" which can then be sent to other users safely.
+For example, if I want to buy a token for 5 MOB, I could create an output which gives me that token, and sign a contingent input which spends 5 of my MOB
+in exchange for someone producing that output. I can then bundle this up and send it to a counter party -- if they agree to trade at my price, they can build
+a transaction using my transaction fragment and submit it.
+
+There is no way that they can spend my 5 MOB without producing the desired output for me, so my funds
+are safe, and moreover, never leave my custody during this process until the moment the trade is final.
```
sugargoat, line 27: 👏

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+requiring each party to "commit" their funds first, then trigger the swap, and have contingencies for each party to recover their committed funds
+if the counter party renegs. These protocols may require several rounds of interaction, and require several on-chain transactions, incurring fees and
+increasing the total delay for the swap to be finished.
+
+This proposal is considerably simpler -- the rounds of interaction are a minimum, requiring only one side to send a transaction fragment to another.
+No one is required to "escrow" their funds into a smart contract or temporary address. A party who desires to swap can broadcast their transaction fragment
+to many potential counterparties. Because the Ring MLSAG contains the key image of the spent input, everyone is able to check if the transaction fragment is
+still valid, and the initiating party can cancel the swap proposal by simply spending input to themselves in a separate transaction, burning its key image and preventing any of the
+broadcasted transaction fragments from being used in a valid transaction.
```
sugargoat, line 49: Maybe worth noting that without escrow, the burden is on the client to provide UX that informs the user how much of their funds are "pending" - this is similar to the way clients work today with transactions. The protocol layer does not report on "pending" or consider funds in a pending state.

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+* A `RingMLSAG` for each `TxIn`
+* A `pseudo_output_commitment` for each `TxIn`
+* A range proof for each `pseudo_output_commitment`, relative to the generator for this token id.
+
+Transaction validation ensures that the sum of outputs minus the sum of pseudo-outputs is the zero point on the elliptic curve, (including the implicit fee output)
+which ensures that the transaction is valid.
+
+This feature contains two discrete steps:
+* Allowing mixed transactions. This requires that individual outputs and pseudo-outputs in a transaction can be labeled with token ids.
+  (The MCIP #25 token id, added to the tx prefix, is now interpretted as the fee token id.)
```
sugargoat, line 72:
```suggestion
  (The MCIP #25 token id, added to the tx prefix, is now interpreted as the fee token id.)
```

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+  When validating range proofs, each individual output and pseudo-output has its range proof checked relative to its specified token id.
+  No other changes are required to support mixed transactions.
```
sugargoat, line 74: And the privacy properties of the on-chain TxOuts also don't change, correct? As this is all in the TxPrefix, which, as discussed in MCIP #25, is delivered through an attested channel and is only visible by the enclave?

garbageslam: Yup!

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+* A `RingMLSAG` signing this.
+* A token id, value, and blinding factor for the `pseudo_output_commitment`.
+* A token id, value, and blinding factor for any output in the `ContingentInputRules`.
+
+Anyone who owns a `TxOut` can use it to create a `SignedContingentInput`. The signed contingent input can be thought of as an offer
+to trade "this for that" -- you are willing to give away your `TxOut` if someone produces for you the `TxOut` that you specify.
+Once you sign the contingent input, you can give it away, and anyone who has it can fulfill your request to trade.
+
+Giving away the signed contingent input does not reveal your public address, nor which `TxOut` in the ledger you actually own.
+It does reveal the token id and value of your pseudo-output, but this is unavoidable in any such scheme.
```
sugargoat, line 93: Can you describe a bit more here? Or maybe this will go in the reference-level explanation - I'm trying to imagine the "shared secrets" involved

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+which is optimal.
+
+Privacy is also a key consideration. In this proposal, the first party sends only a `SignedContingentInput`,
+so the second party does not learn their public address, only what they propose to trade and for how much.
+The first party does not learn anything about the second participant except that they executed the trade.
+The second party is unable to send subsequent transactions to the first party, since all they got was an output
+created by the first party, and the TxOut `public_key` is enforced to be unique across the ledger.
+
+In a realistic situation, it seems likely that the first and second parties know who each other are, since
+they must initiate the interaction somehow. However, that is not required by the protocol proposed here.
```
sugargoat, line 126: Do you imagine some evolution where there could be "contingent input pools?" Not necessarily on MobileCoin's layer1 quite yet, but this seems like a great use case

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,191 @@
...
+
+Additionally, because the first party does not send a message to the blockchain to initiate the trade,
+and only sends a `SignedContingentInput` instead, there is no permanent record that the trade was proposed,
+nor is there a fee for proposing the trade. Additionally there is no possibility of "Miner Extractable Value"
```
sugargoat, line 130: 👏

text/0031-transactions-with-contingent-inputs.md
garbageslam: Yeah it seems quite plausible to me, like a `SignedContingentInput` is effectively like a limit order in an order book I think, except that there's no custodian and anyone who has the order is capable of executing it if they want. So you could conceivably have like a distributed order book of inputs like this, transmitted via a gossip network or something, and any participant is capable of executing any of the orders.

I think it would also be possible to merge lots of orders into one transaction and execute them all at once.

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,394 @@
...
+In a MobileCoin transaction, an input consists of a collection of `TxOut`'s in the ledger, one of which is truly being spent and others of which are just being mixed in. This ring,
+and the proofs of membership, appear in the `TxIn`, in the `TxPrefix`. A transaction produces several new `TxOut`'s, and these outputs appear in the `TxPrefix`.
+
+The signature part of a transaction contains several things:
+* A `RingMLSAG` for each `TxIn`
+* A `pseudo_output_commitment` for each `TxIn`
+* A range proof for each `pseudo_output_commitment`, relative to the generator for this token id.
+
+Transaction validation ensures that the sum of outputs minus the sum of pseudo-outputs is the zero point on the elliptic curve, (including the implicit fee output)
+which ensures that the transaction is valid.
```
iamalwaysuncomfortable, line 75: If I recall my Risretto math correctly, there are different prime subgroups which a token might belong to right - do each token types inputs and outputs independently sum to the zero point? (Say if you have MoB and MeowB the MoB-only inputs/outputs sum to the zero point & the MeowB inputs + ouputs independently sum to the zero point?)

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,394 @@
...
+and another user fills in the rest of the transaction and signs the other MLSAGs.
+
+Therefore we propose that when signing `RingMLSAG`'s, there are two different cases, one for the situation where there are `ContingentInputRules`,
+and the other for the case where there are not `ContingentInputRules`.
+* When there are no `ContingentInputRules` on the input, the message to the MLSAG is the `extended_message_digest` the same as before: https://github.com/mobilecoinfoundation/mobilecoin/blob/694fd2de00443338e6dfc63359c6178840cf7e52/transaction/core/src/ring_signature/rct_bulletproofs.rs#L209
+* When there are contingent input rules present on the input, the message signed by the the MLSAG just consists of the digest of the `TxIn` but always omitting the `TxOutMembershipProof`, and including any `ContingentInputRules`.
+
+(The `TxOutMembershipProof` should not be signed because consensus may enforce that membership proofs have similar root membership elements. There is no problem to allow the final submitter of the
+transaction to update the root membership elements of these proofs.)
```
iamalwaysuncomfortable, line 203: There seems to be a lot of checking that we should maybe be doing here, is it possible to quickly sketch the math of this out in latex or in another form that could be checked?

iamalwaysuncomfortable (approved):
This is great because it would create quite the token ecosystem out of MoB if atomic swaps are supported natively on chian. I'd advocate for a lot more study of this though and perhaps a formal draw up of the mathematics for checking & experimentation.

text/0031-transactions-with-contingent-inputs.md
garbageslam: Yeah, so, suppose G_1 is the base for MOB and G_2 is the base for MEOWB, and H is the blinding factor.

The transaction math checks that the sum of outputs + fee output - sum of pseudo outputs is equal to 0, and actually, the zero point of the elliptic curve:

https://github.com/mobilecoinfoundation/mobilecoin/blob/acbf10842b0d17a22cce0a2123886f03bebea487/transaction/core/src/ring_signature/rct_bulletproofs.rs#L195

So effectively, each output and pseudooutput that we add here is a vector Pedersen commitment, so this is like

(output_1_value * G_1 + blinding * H) + (output_2_value * G_2 + blinding * H) + (fee_value * G_1 + blinding * H) - (input_1_value * G_1 + blinding * H) - (input_2_value * G_2 + blinding * H) = 0

If we collect terms, this is like

(MOB_balance * G_1) + (MEOWB_balance * G_2) + (balance_of_blinding_factors * H) = 0

At this point, we want to use the idea that the generators are all orthogonal -- it is intractable for any person to find a nontrivial linear relation between `G_1, G_2, H`. This is happening because they were hashed to the curve, and decision diffie hellman impllies that random curve points are orthogonal. (also mentioned in comments on MCIP 25)

So, if the person who built the transaction can't find a nontrivial linear relation, then to satisfy the equation they must have made a trivial linear relation. So

* `MOB_balance = 0`
* `MEOWB_balance = 0`
* `balance_of_blinding_factors = 0`

So this is the "vector pedersen commitment" idea (also in bulletproofs), where like, we can basically think of `G_1, G_2, H` as a basis for some vectorspace. So this is why its okay to have mixed inputs and outputs in the balance check part now.

The only reason that the token ids even need to be revealed to the enclave here, is that we have to check the validity of the range proofs, and we can't do that if we don't know the generator for that token.

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,394 @@
...
+
+This creates the possibility that the extended message digest is never signed -- if every `TxIn` has contingent input rules, then none of them will actually sign over the `extended_message_digest`,
+and if there are additional outputs in the `Tx` besides those appearing in the rules, no one will have signed over them, and no one will have signed over the fee. This would cause these parts of the
+transaction to become malleable -- if the process of proposing the Tx to consensus, a man in the middle were able to intercept it, they could tamper with these parts of the transaction, in a way that
+isn't possible for transactions currently.
+
+Since transactions are submitted over attested channels to the consensus enclave, there is already quite a lot of protection against tampering with in-flight transactions.
+However, another simple way to mitigate this is that a client can ensure that at least one input to the `Tx` does not have contingent input rules, and so signs the whole `TxPrefix`.
```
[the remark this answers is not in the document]
garbageslam: Yeah, we could potentially make it a protocol rule as well, if we wanted, that there has to be at least one "normal" input

text/0031-transactions-with-contingent-inputs.md
garbageslam: I am happy to try to add a section where we try to articulate the security property precisely. I would like to avoid adding a companion latex document if possible, if only because that's not currently part of the MCIP process. We could explore making tex files also part of the MCIP repo I guess.

text/0031-transactions-with-contingent-inputs.md
iamalwaysuncomfortable: Thanks for explaining this!

text/0031-transactions-with-contingent-inputs.md
```diff
@@ -0,0 +1,408 @@
...
+    repeated external.TxOut required_outputs = 1;
+    // A maximum tombstone block value for the transaction. If zero, it means no maximum is enforced.
+    fixed64 max_tombstone_block = 2;
+}
+```
+
+It may be hashed in a standard way using the `mc-crypto-digestible` hashing scheme (like everything else).
+
+This `required_outputs` are there to allow the main use-case of agreeing to a swap.
+The `max_tombstone_block` is there both, to allow that some of the required outputs can be fog outputs (since fog imposes tombstone block limits),
```
wjuan-mob, line 176:
```suggestion
The `max_tombstone_block` is there both to allow that some of the required outputs can be fog outputs (since fog imposes tombstone block limits),
```

wjuan-mob (approved):

garbageslam:
This proposal has now progressed to the point where we have a working POC and have successfully submitted atomic swap transactions on a dev network. At this point I think we should decide whether to accept it the proposal and not increase it's scope further.

Moving to final comment period

PR #32, opened 2022-03-21
Fog-compatible Gift Codes

garbageslam:
A proposal for a different mechanism to implement the gift codes feature, which is simpler and compatible with Fog.

[Rendered Proposal](https://github.com/garbageslam/mcips/blob/fog-compatible-gift-codes/text/0032-fog-compatible-gift-codes.md)

### DIFF
--- a/text/0032-fog-compatible-gift-codes.md
+++ b/text/0032-fog-compatible-gift-codes.md
@@ -41,8 +41,10 @@ However, there could be several deployments of Fog, and this basically means tha
 what software the recipient is using, which is counter to the goal of the feature.
 
 We propose to fix this in a quite simple way:
+
 Instead of transfering private keys to an entire account, we just transfer ownership of a specific `TxOut`.
-We give them private keys and secrets connected to this `TxOut`, but which don't reveal anything about our account private 
+We give them private keys and secrets connected to this `TxOut`, but which don't reveal anything about our account private keys.
+
 * To enable someone to spend a TxOut, we just have to give them the `onetime_private_key` of the `TxOut`.
 * To enable someone to see the unblinded amount of the TxOut, we just have to give them the `shared_secret` of the `TxOut`.
 * To enable someone to find the TxOut in the ledger (even a Fog user), we just have to tell them the global index in the blockchain.
--- a/text/0032-fog-compatible-gift-codes.md
+++ b/text/0032-fog-compatible-gift-codes.md
@@ -219,18 +219,18 @@ The memo on the change output is a note Alice writes to herself to keep track of
 memos by confirming that they belong to the change subaddress. The memo on the other output is a note from Alice to Bob,
 typically identifying Alice.
 
-When funding a gift code, the TxOut to `GIFT_CODE_SUBADDRESS` if the gift code tx out, and a change output is always written (even if value zero):
+When building a transaction which funds a gift code, the TxOut to `GIFT_CODE_SUBADDRESS` is called the "gift code output", and a change output is always written (even if value zero):
 * `0x0201 Gift code funding memo` should be attached to the change output. This memo will point to the gift code output by way of a hash, and contains any note about the purpose of the gift code.
 * `0x0001 Unauthenticated sender memo`, or `0x0002 Unauthenticated sender memo with address hash`, is attached to the gift code output.
 
-When claiming a gift code, the gift code output is spent in a self-payment to the change subaddress. The transaction has a single output, the
-change output, which goes to the change subaddress, and has a memo of type `0x0001 Unauthenticated sender memo`, or `0x0002 Unauthenticated sender memo with address hash`.
+When claiming a gift code, the gift code output is spent in a self-payment to the change subaddress. This transaction has a single output, the
+change output, which has a memo of type `0x0001 Unauthenticated sender memo`, or `0x0002 Unauthenticated sender memo with address hash`.
 The app should attempt to validate the information in the memo from the gift code -- for example, in a chat application, the app may be able
-to validate who sent the gift code message. If the information is valid, then the app should copy the memo from the gift code TxOut into the
+to validate who sent the gift code payload in a message. If the information is valid, then the app should copy the memo from the gift code TxOut into the
 self-payment TxOut. If the app decides that the information cannot be validated, then it may use `0x0000 Unused` instead, to avoid committing
 false information. Future apps may assume that any memos attached to TxOut's owned by the change subaddress are known to be valid.
 
-When cancelling a gift code, the gift code tx out is paid back to self. The transaction has a single output, the change output, which goes to
+When cancelling a gift code, the gift code output is paid back to self. This transaction has a single output, the change output, which goes to
 the change subaddress and has a `0x0202 Gift code cancellation memo`. This memo contains the tx public key of the gift code output.
 
 The overall strategy for an app implementing recoverable transaction history looks like this then:
@@ -336,11 +336,6 @@ None that we are aware of.
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
-It seems that we are often in a position of reserving subaddresses for new purposes discovered in mobile applications.
-
-This conflicts with how subaddresses are used in applications like `mobilecoind` and `full-service`, which are often assigning
-a new subaddress per contact.
-
-Since we want for users to be able to export their account from mobile apps and load them in desktop apps and have things work cleanly,
-it may be a good idea to reserve a specific range of subaddresses for mobile, so that when a desktop app loads account history it doesn't
-get confused by the way mobile has been using the subaddresses.
+In the future we may want to extend the gift code memos, e.g. to include payment request ids or something. But this set of memos seems to meet
+the goals at end, including, creating memos for a peer, tracking, and cancelling them, and recovering this history, as well as creating gift
+codes with text notes.
--- a/text/0032-fog-compatible-gift-codes.md
+++ b/text/0032-fog-compatible-gift-codes.md
@@ -15,7 +15,7 @@ compared to the existing mechanism.
 # Motivation
 [motivation]: #motivation
 
-Full service supports a "gift codes" feature. (Called "Tranfer code" [in sources](https://github.com/mobilecoinfoundation/mobilecoin/blob/7077b418fab65e05a05e4fca1d272e3dddd2e392/api/proto/printable.proto#L28).)
+Full service supports a "gift codes" feature. (Called "Transfer code" [in sources](https://github.com/mobilecoinfoundation/mobilecoin/blob/7077b418fab65e05a05e4fca1d272e3dddd2e392/api/proto/printable.proto#L28).)
 
 The idea of a gift code is that you can give money to
 someone without knowing their public address before hand, creating a payload that can be handed to
--- a/text/0032-fog-compatible-gift-codes.md
+++ b/text/0032-fog-compatible-gift-codes.md
@@ -15,7 +15,7 @@ simpler since it doesn't require creating a temporary account, and instead gives
 directly.
 
 We propose that TxOut's which are sent as gift codes, should be paid first to the reserved
-`GIFT_CODE_SUBADDRESS_INDEX`, which is equal to `u64::MAX - 1` (per MCIP #36).
+`GIFT_CODE_SUBADDRESS_INDEX`, which is equal to `u64::MAX - 2` (per MCIP #36).
 
 # Motivation
 [motivation]: #motivation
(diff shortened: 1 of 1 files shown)


the-real-adammork (commented):
so if I send a the credentials for one of my TxOuts as a `TxOutGiftCode`, theres a chance the receiver wont be able to redeem because that TxOut gets spent by the sender. We'll have to manage that on the clients ?

garbageslam:
> so if I send a the credentials for one of my TxOuts as a TxOutGiftCode, theres a chance the receiver wont be able to redeem because that TxOut gets spent by the sender. We'll have to manage that on the clients ?

Thank you, this is a good question that I had not fully considered.

Context for other readers: For clients like signal, which support linked devices, the account credentials may be in multiple places, and they cannot all be fully synchronized. (Some of them may not be even be reachable for instance.)

This basically means that whenever the phones send money from one place to another, it is racy, and if two linked devices try to build transactions, it's possible they select the same inputs and one of the transactions fails.

With the existing transfer payload, if I we both have phones and I want to send a transfer payload to you, first I fund it, which is racy between my clients, but with retries it eventually either succeeds, or my balance isn't enough. Then I can send you the funded transfer payload, and your clients will be trying to transfer immediately to their account. They will be racing eachother, but not be racing mine (since I am trying to send you the money and not drain the funded account, only one of my clients even knows that address).

With this version, there is no transaction that commits funds the transfer code, I'm just sending you private keys to my `TxOut`. But my other phones don't know this happened and may also be racing to spend that `TxOut`. The difference now is that it's possible that I think I sent you a funded transfer code, but my other device wins the race for you to use it to fund your account, and it may even be difficult for me to be sure that that is what happened.

I also won't really have any transaction history around the gift code, which is negative.
(Note: We should perhaps create some recoverable transaction history memo which is appropriate for funding and redeeming gift codes, I think right now it would always just show as an unknown account.)

---

So, maybe this isn't really better if the sender may have linked devices. Maybe it is better to have the transfer account in the middle, to resolve races between sender and recipient and to ensure that the sender has transaction history, even though it increases end-to-end time and costs an extra fee.

In some cases the sender may be able to ensure that they don't have linked devices, and in this case I think this version is preferable. But I'm not sure if that scenario is more important.

---

I still think it would be better if the transfer code has the global tx out index instead of the tx public key. But that's a much smaller change.

Thanks for your comments @the-real-adammork !

the-real-adammork:
So I guess Im thinking of a situation where I send you a "gift code" and it doesn't get redeemed right away, say that time period is non-trivial like an hour or a day. During this time if I want to make a new transaction the SDK will need to exclude that TxOut from balance & transaction building, but where do I store this state ?

With the existing solution, Im funding another account so my balance changes immediately, I don't have to manage any state. It can be redeemed by the receiver party whenever they have time, or by me in the future if I still have the credentials and it was never redeemed.

Is there a hybrid approach where we can move this TxOut somewhere or annotate it with some state so the SDKs can ignore it in balance and transaction building ? Maybe we add a new MemoBuilder type for this purpose ?

mfaulk:
This is a big improvement over the earlier gift code design. 

The earlier design, where a "gift" contains account private keys, is really a terrible design! The point of private keys is that they should be private, so it seems like asking for trouble to create scenarios where multiple people (i.e. the giver and receiver of the account keys) know the account keys. Imagine a scenario where a malicious "giver" monitors the gift account; a naive recipient loads the gift's account keys into some software to "receive" the gift, but then later uses that account's public keys to receive other MOB. The malicious giver sees this because he also has the account private keys, and immediately transfers that MOB to his own account.

For my $0.02, I don't think we should implement gift codes. They seem like a lot of hassle and potential problems just to implement an atypical user experience.

the-real-adammork:
> @the-real-adammork I suppose we could do a two-step thing, where the sender creates a temporary account, funds that account by putting a TxOut on the ledger, then puts the TxOut secrets into the gift code. The sender doesn't need to keep track of whether that TxOut has been spent, and the recipient is never given "shared" private account keys.

I think this is a usable solution too, but I like the idea of me owning the TxOut until its redeemed. I think in the case where a gift-code goes un-redeemed and you want to redeem it, you can do that without store any additional information.

text/0032-fog-compatible-gift-codes.md
```diff
@@ -0,0 +1,201 @@
...
+They can get the entire `TxOut` this way, match it against the shared secret and so unblind the amount commitment, and given the one-time
+private key, they can spend it as normal using the transaction builder.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A `TxOutGiftCode` is a payload which confers ownership of an individual `TxOut` in the blockchain,
+and is a little less than 80 bytes in size.
+
+* To enable someone to spend a TxOut, we just have to give them the `onetime_private_key` of the `TxOut`.
```
iamalwaysuncomfortable, line 92: As cited elsewhere, there's a race to spend. Notably a lot of clients tend to select txos based on optimization based algos. Clients would have to implement some kind txo locking mechanism with a ridiculously sized tombstone block (maybe like as in real gift cards - it has a redemption window) see: https://github.com/mobilecoinofficial/full-service/pull/270 in order to implement this in a way that's safe. 

There's a small enough amount of client languages I think where it *might* be reasonable for us to provide a set of traits/interfaces (java/android)/protocols (swift/ios) like "CanLockTxo" for clients that they'd need to implement to be able to use this feature. It could be enforced in some cases for things that use the specific APIs we provide. For Signal it would be feasible for us to PR these into their client to provide this functionality. 

Question is: is that work worth it?

voloshyn: You'd have to share the information about the "reserved" TxOuts between the linked clients, while you could do it on Signal via the message channel, other clients though, unfortunately, can't do it. Hence we can't have a guarantee that any other client that has access to this account won't accidentally spend your gift code before it was redeemed.

voloshyn:
> @the-real-adammork I suppose we could do a two-step thing, where the sender creates a temporary account, funds that account by putting a TxOut on the ledger, then puts the TxOut secrets into the gift code. The sender doesn't need to keep track of whether that TxOut has been spent, and the recipient is never given "shared" private account keys.

This does not seem much different from the existing `TransaferPayload`. Both sender and recipient can still spend that TxOut because they have "_shared_" private keys. Using `TransaferPayload` we can derive `onetime_private_key` and `shared_secret` from the `AccountKey` + `TxOut` combo provided in `TransaferPayload`.
So it seems like the only use case that will be solved is when the recipient tries to fund a used gift-code account. It's technically possible but very unlikely to happen since it would require developers deliberately develop this functionality to even enable such a use case.
Also, the ability to cancel gift-card payments seems like a feature to me, not a bug. Moreover, it was a requirement from the product during our previous iteration.

voloshyn (commented):
Alternatively, instead of creating a temporary account, we could use a dedicated subaddress for all `TxOutGiftCode`s so: 
* any client could see completed and pending gift codes without having to sync up with the original client.
* clients that don't support gift codes functionality will still show the correct balance.
* shared private keys are valid for only 1 TxOut and cannot be re-used.

text/0032-fog-compatible-gift-codes.md
```diff
@@ -0,0 +1,201 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+In our opinion, there is no reason why we should not do this, and this version of gift codes is superior to the existing version.
+The existing version is very complicated and forces all the clients to view-key scan the ledger with a new set of keys in order
+to process a gift code. The "fixed" version that returns the `TxOut` public key avoids the view key scanning issue, but it still
+isn't really compatible with Fog's privacy model. It is much simpler to just give away ownership of an existing `TxOut`, and
+give the fog recipient a global `TxOut` index.
+
+This also makes creating gift codes instantaneous and not require on-chain events and fees.
```
voloshyn, line 179: This is true only if you have a TxOut with an exact amount, i.e. If Alice wants to send Bob 5MOB, but only has a single TxOut of value 10 MOB, she will have to first send 5 MOB to her own PublicAddress.

garbageslam:
Thank you @voloshyn i'm going to incorporate your suggestions into the MCIP proposal

voloshyn (approved):

text/0032-fog-compatible-gift-codes.md
```diff
@@ -0,0 +1,255 @@
...
+  * To enable a recipient to spend a `TxOut`, we just have to give them the `onetime_private_key`.
+  
+Given this data, a Fog user who receives the payload can ask Fog ledger for a merkle proof of membership for the `TxOut` of given index.
+They can get the entire `TxOut` this way, match it against the shared secret and so unblind the amount commitment, and given the one-time
+private key, they can spend it as normal using the transaction builder.
+
+A major concern for mobile apps is the ability to track things across linked devices using the blockchain (and not a separate synchronization
+mechanism). It is a concern for instance that one app may be trying to give away a `TxOut` in a gift code, while another selects it to be spent in a transaction.
+
+To avoid that, we propose to standardize that `GIFT_CODE_SUBADDRESS_INDEX = 2`, and that before building a gift code, the app should make a self-payment to this
```
eranrund, line 97:
```suggestion
To avoid that, we propose to standardize that `GIFT_CODE_SUBADDRESS_INDEX = u64::MAX - 1`, and that before building a gift code, the app should make a self-payment to this
```

text/0032-fog-compatible-gift-codes.md
```diff
@@ -0,0 +1,255 @@
...
+subaddress witht he correct amount for the gift code. Then the gift code is built against the `TxOut` that belongs to subaddress 2.
```
eranrund, line 98:
```suggestion
subaddress with the correct amount for the gift code. Then the gift code is built against the `TxOut` that belongs to subaddress `u64::MAX - 1`.
```

text/0032-fog-compatible-gift-codes.md
```diff
@@ -0,0 +1,255 @@
...
+
+The advantage of this is that unlike with the temporary address, the fog user is able to track the gift code.
+* The `TxOut` still belongs to the sender's address, even if we gave the private keys away to someone (who may not have used it yet).
+* We don't lose track of the in-flight gift code if the app is restarted -- all linked devices can learn about the in-flight gift code because of the use of subaddress 2.
```
eranrund, line 102:
```suggestion
* We don't lose track of the in-flight gift code if the app is restarted -- all linked devices can learn about the in-flight gift code because of the use of subaddress `u64::MAX - 1`.
```

text/0032-fog-compatible-gift-codes.md
```diff
@@ -0,0 +1,255 @@
...
+  This was not the case with the temporary address.
+* Because the `TxOut` still belongs to us, we can see if the gift code is actually used by the recipient, and if for some reason it is never used, we can
+  reclaim the money. (This would still be possible with the previous design, but it's more complicated because we have to preserve the temporary address entropy off-chain somewhere.)
+* The app can also more simply have recoverable transaction history around the use of gift codes, because they can represent in-bound `TxOut`'s to subaddress 2 as gift-code funding
```
eranrund, line 106:
```suggestion
* The app can also more simply have recoverable transaction history around the use of gift codes, because they can represent in-bound `TxOut`'s to subaddress `u64::MAX - 1` as gift-code funding
```

garbageslam:
we have to update the subaddress again to `-2`

text/0032-fog-compatible-gift-codes.md
```diff
@@ -0,0 +1,255 @@
...
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+It seems that we are often in a position of reserving subaddresses for new purposes discovered in mobile applications.
+
+This conflicts with how subaddresses are used in applications like `mobilecoind` and `full-service`, which are often assigning
+a new subaddress per contact.
+
+Since we want for users to be able to export their account from mobile apps and load them in desktop apps and have things work cleanly,
```
nick-mobilecoin, line 253: Not sure how much we want to try and keep _all_ MCIPs up to date, especially when multiple associated ones end up in flight at the same time.
This could reference MCIP [36](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0036-reserved-subaddresses.md) or perhaps this future possibilities section should go away.

I'm also fine with leaving this as is if that is desired.

nick-mobilecoin (approved):

text/0032-fog-compatible-gift-codes.md
garbageslam: FWIW I think we should try to keep all the MCIPs up to date, so that readers who jump into our ecosystem can understand what's going on

nick-mobilecoin: Given that I would vote for removing the future work section from this MCIP since it has now been accepted as [MCIP 36](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0036-reserved-subaddresses.md)

garbageslam: thank you

eranrund (approved):

PR #35, opened 2022-04-07
MCIP for verifiable burn address

garbageslam:


text/0035-verifiable-burning.md
```diff
@@ -0,0 +1,165 @@
...
+
+To provably burn funds, simply send them to the burn address. You may attach a memo if this serves your purposes.
+
+To verify that burning occurred, use the view private key to the burn address and view-key scan the entire blockchain.
+Every provably burning event will be revealed, and you can see the block index, the amount, and read the memo field.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The burn address construction and rationale is as follows:
```
nick-mobilecoin, line 50: What prevents *anyone* from burning arbitrarily via sending funds to the burn address?

garbageslam: Nothing, anyone is welcome to burn to the burn address. If some sort of authentication e.g. for integration with a redemption workflow, then I'd want that to be like a Mac value in the encrypted memo or something like this

text/0035-verifiable-burning.md
```diff
@@ -0,0 +1,165 @@
...
+# Motivation
+[motivation]: #motivation
+
+Burning means destroying funds (or, rendering them unspendable).
+
+There are a few reasons this may be done:
+* To adjust the supply of a synthetic asset
+* To create a deflationary pressure on a token
+
+Particularly, [MCIP #31](https://github.com/mobilecoinfoundation/mcips/pull/0031)
```
garbageslam, line 20:
```suggestion
Particularly, [MCIP #25](https://github.com/mobilecoinfoundation/mcips/pull/0025)
```

nick-mobilecoin (approved):

wjuan-mob (commented):
I understand the argument that given a hash function, the hardness is equivalent to solving discrete log. But is there a possibility that people will distrust our choice of hash function? Given a choice of any hash function, could we have picked a hash function that for this value particularly yields a private key that we can know?

garbageslam:
> Given a choice of any hash function, could we have picked a hash function that for this value particularly yields a private key that we can know?

Yes -- we have to assume that blake2b is a trustworthy hash function which has not been manipulated. But fortunately, it is widely used and very well audited.

text/0035-verifiable-burning.md
```diff
@@ -0,0 +1,165 @@
...
+fn burn_address_view_public() -> RistrettoPoint {
+    BURN_ADDRESS_VIEW_PRIVATE * burn_address_spend_public()
+}
+```
+
+That is, the product of account view private key, with subaddress spend public key,
+is equal to the subaddress view public key.
+
+This identity can be verified by inspecting the MobileCoin subaddress derivation implementation,
+or by referring to Mechanics of MobileCoin.
```
sugargoat, line 92:
```suggestion
or by referring to [Mechanics of MobileCoin](https://github.com/mobilecoinfoundation/Mechanics-of-MobileCoin).
```

text/0035-verifiable-burning.md
```diff
@@ -0,0 +1,165 @@
...
+  for discrete log in curve25519.
+
+Thus, we have shown by reduction to discrete log that it is intractable to find the subaddress spend key for these funds.
+
+This security argument also shows why it isn't important for the view private key to be chosen by hashing.
+
+# Prior art
+[prior-art]: #prior-art
+
+None at this time.
```
sugargoat, line 153: You could link to the previous proof-of-concept implementations [here](https://github.com/christian-oudard/mc_incinerator) and [here](https://github.com/mobilecoinfoundation/mobilecoin/pull/1581)?

sugargoat (approved):
Looks great!

text/0035-verifiable-burning.md
```diff
@@ -0,0 +1,165 @@
...
+The `BURN_ADDRESS_VIEW_PRIVATE_KEY` is a curve25519 scalar, and is a constant known to everyone.
+
+```
+pub const BURN_ADDRESS_VIEW_PRIVATE: Scalar = Scalar::from_bits([1u8; 32]);
+```
+
+The bit pattern `[1u8; 32]` was chosen arbitrarily. It is the next simplest bit pattern to write in code after `[0u8; 32]`.
+We did not choose to make this scalar zero, because, zero is the "funniest number in cryptography" and sometimes leads to degeneracies.
+
+The `burn_address_spend_public` key is a ristretto curve point, intended to be a "nothing up my sleeve" number,
```
isis-mc, line 61: small nitpick: s/ristretto curve point/ristretto element/

(Ristretto is a prime-order group, abstracted over families of potential elliptic curves, the elliptic curves themselves being implementation details. There isn't a curve equation per se to define Ristretto, as there is with a twisted Edwards curve like curve25519 or a pseudo-Koblitz curve like secp256k1.)

isis-mc (approved):

text/0035-verifiable-burning.md
garbageslam: thank you --

I feel like, I really want to follow the existing docu that we already have:
* All the code says `RistrettoPoint` because the type curve25519-dalek crate exposes is called `RistrettoPoint`
* Mechanics of MobileCoin also uses the term "Ristretto abstraction" rather than Ristretto curve or Ristretto group, but still calls the elements "Ristretto Point" rather than "Ristretto Element"

I'm going to try to use the term Ristretto point rather than Ristretto curve point

PR #36, opened 2022-04-09
Reserved Subaddresses

garbageslam:
A proposal to reserve a range of subaddresses for future purposes, to promote compatibility.

[Rendered Proposal](https://github.com/garbageslam/mcips/blob/reserved-subaddresses/text/0036-reserved-subaddresses.md)

### DIFF
--- a/text/0004-recoverable-transaction-history.md
+++ b/text/0004-recoverable-transaction-history.md
@@ -117,7 +117,7 @@ that went to the change subaddress.
 ## Change subaddress
 
 The "change subaddress" is a new standard subaddress, like the default subaddress. It should be used when sending change outputs.
-It is specified to be subaddress index `u64::MAX`.
+It is specified to be subaddress index `u64::MAX-1`.
 
 The change subaddress should be kept secret, so that only authentic 0x0200 Destination memos will pass validation.
 
--- a/text/0004-recoverable-transaction-history.md
+++ b/text/0004-recoverable-transaction-history.md
@@ -23,7 +23,7 @@ history given only the private keys of the user, by fetching all of the user's T
 decoding and validating the memos.
 
 Some additional ancillary parts of the proposal:
-- Change subaddress is standardized to subaddress index `u64::MAX`.
+- Change subaddress is standardized to subaddress index `u64::MAX - 1`.
 - Clients should always produce a change output even if the change is zero, in order to have a place to put
   the destination memo.
 - A standard 16-byte hash of a public address is specified, which may be useful elsewhere.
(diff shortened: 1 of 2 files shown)


text/0036-reserved-subaddresses.md
```diff
@@ -0,0 +1,68 @@
...
+
+Subaddress indices are unsigned 64-bit numbers.
+
+The indices with high bit set are reserved for "special purposes", and should
+not be assigned to individual contacts by desktop wallets that are using subaddresses this way.
+
+`u64::MAX` will be reserved for an "invalid / none" value.
+
+[mobilecoinfoundation/mcips#4](https://github.com/mobilecoinfoundation/mcips/pull/4) will be updated,
+changing the change subaddress from `1` to `u64::MAX - 1`.
```
sugargoat, line 16: Why would we change the change subaddress?

sugargoat: It seems like a high cost to change, as this has been deployed as a subaddress for change for more than a year

sugargoat: Just synced with @garbageslam offline, and here's where full-service is setting it:

https://github.com/mobilecoinofficial/full-service/blob/main/full-service/src/db/account.rs#L25

the-real-adammork (approved):

text/0004-recoverable-transaction-history.md
```diff
@@ -115,7 +117,7 @@ that went to the change subaddress.
 ## Change subaddress
 
 The "change subaddress" is a new standard subaddress, like the default subaddress. It should be used when sending change outputs.
-It is specified to be subaddress index 1.
+It is specified to be subaddress index `u64::MAX`.
```
dolanbernard, line 120: -1

dolanbernard (changes requested):

text/0004-recoverable-transaction-history.md
the-real-adammork, line 120:
```suggestion
It is specified to be subaddress index `u64::MAX-1`.
```

@dolanbernard you can propose the change in the comment with the +/- "document" icon

the-real-adammork: nice catch tho

text/0036-reserved-subaddresses.md
garbageslam: @sugargoat the problem is that, mobilecoind was already using this subaddress for customers, and they were in the wild even before full-service.

the plan right now is that we would like for mobilecoind users to move to full-service. But it's impossible for them to do that if full-service is treating this subaddress as change. 

* They cannot claw back subaddresses that they already gave to users, which users might continue to deposit money into
* If the change subaddress is known by anyone, then that person is able to forge destination memos in the RTH scheme. So it is very unsafe for an exchange to start using subaddress 1 as a change address.

Since it is security critical that the change subaddress is not known by anyone, and some clients were using subaddress 1 casually, the only safe option is to use a completely new subaddress as the change subaddress.

text/0004-recoverable-transaction-history.md
garbageslam: thank you

text/0036-reserved-subaddresses.md
garbageslam: we can mention this in the drawbacks / rationale section

dolanbernard (approved):

remoun (approved):

PR #39, opened 2022-05-02
Burn Redemption memo

eranrund:
A proposal to introduce a new RTH memo type for storing unstructured information associated with burn transactions.

[Rendered Proposal](https://github.com/eranrund/mcips/blob/burn-redemption-memo/text/0039-burn-redemption-memo.md)

### DIFF
--- a/text/0039-burn-redemption-memo.md
+++ b/text/0039-burn-redemption-memo.md
@@ -31,7 +31,7 @@ This is a simple proposal. We propose reserving the memo type 0x0001 for the bur
 | Memo type bytes | Name                                              |
 | -----------     | -----------                                       |
 | 0x0001          | Burn redemption memo                              |
-|
+
 
 ## 0x0001 Burn redemption memo
 
--- a/text/0039-burn-redemption-memo.md
+++ b/text/0039-burn-redemption-memo.md
@@ -66,9 +66,9 @@ None that we are aware of.
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-# Future possibilities
-
 None that we are aware of.
+
+# Future possibilities
 
 [future-possibilities]: #future-possibilities
 
--- a/text/0039-burn-redemption-memo.md
+++ b/text/0039-burn-redemption-memo.md
@@ -1,6 +1,6 @@
 - Feature Name: Burn-Redemption-Memo
 - Start Date: 2022-05-02
-- MCIP PR: [mobilecoinfoundation/mcips#0000](https://github.com/mobilecoinfoundation/mcips/pull/0000)
+- MCIP PR: [mobilecoinfoundation/mcips#39](https://github.com/mobilecoinfoundation/mcips/pull/39)
 - Tracking Issue: N/A
 
 # Summary
--- a/text/0000-burn-redemption-memo.md
+++ b/text/0000-burn-redemption-memo.md
@@ -1,0 +1,75 @@
+- Feature Name: Burn-Redemption-Memo
+- Start Date: 2022-05-02
+- MCIP PR: [mobilecoinfoundation/mcips#0000](https://github.com/mobilecoinfoundation/mcips/pull/0000)
+- Tracking Issue: N/A
+
+# Summary
+[summary]: #summary
+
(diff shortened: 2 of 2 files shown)


text/0039-burn-redemption-memo.md
```diff
@@ -0,0 +1,75 @@
...
+| Memo type bytes | Name                                              |
+| -----------     | -----------                                       |
+| 0x0001          | Burn redemption memo                              |
+
+
+## 0x0001 Burn redemption memo
+
+| Byte range | Item |
+| ---------- | ---- |
+| 0 - 64     | Unstructured payload  |
```
iamalwaysuncomfortable, line 40: This all seems perfectly logical to make completely unstructured so that we're not making assumptions. If we need something more, we can add something more.

However, does it make sense for us to possible reserve memo types that might fall under the classification of "burn tx" (similar to how 0x0200 are change memos & 0x0100 are destination memos) that can be diversified under the superclass of "burn memos"? It could feasibly in the future need more structure due to excessive fragmentation or regulation pressure. Say we need a "chain identifier + burn tx indentifier" or something, is there an address space of memo types we can currently reserve for this (say the space under 0x010)?

iamalwaysuncomfortable (commented):
Comments on potentially needing more address space later for burn txs

text/0039-burn-redemption-memo.md
iamalwaysuncomfortable: Gift code memos were reserved at 0x0002 and it's feasible other memo types could just start popping up behind that. We could just implement 0x0003.. 0x0004.. etc. - but maybe given this is very regulation heavy, we create a class for it if we need more space?

text/0039-burn-redemption-memo.md
```diff
@@ -0,0 +1,75 @@
...
+
+While it is possible to add more information to the ledger to store this kind of data, it would make no sense as the majority of `TxOut`s are not burn `TxOut`s. Additionally, the memo field was created exactly for this kind of auxiliary data.
+
+We could enforce a schema over the payload, e.g. the first byte could specify what kind of address or id is stored in the remaining 63 bytes. However, this seems suboptimal since it is impossible for us to predict how this field would get used and we do not want to enforce arbitrarily limits on entities choosing to use this mechanism to track burns. Additionally, there is no way for Consensus to actually verify the contents of the memo - so such schema would be a recommendation at most, but anyone using this memo could still store whatever they desire in the payload and ignore the schema.
+
+We could avoid standardizing this memo field altogether, but as mentioned above, this is not enforceable in the transaction validation code. As such, we think it is better to explicitly reserve a memo type for this, so that it does not get re-purposed by future MCIPs or properly-behaved clients.
```
iamalwaysuncomfortable, line 58: IMHO, it definitely is better and we might consider reserving a small bit of extra space for the burn memos. They could potentially fall under "minting memos"

text/0039-burn-redemption-memo.md
eranrund: I don't think there's a requirement for the memo type bytes to be sequential, so leaving gaps doesn't seem necessary to me but I also don't have any strong preference about that so we could move this to be 0x0010 and leave a few unused ones before it.

iamalwaysuncomfortable: Yeah there's definitely no strict requirement for it. Let's see what other people think

isis-mc (approved):
I'm mildly in favour of keeping the full 64 bytes as unstructured data, since that would allow us to do things like store a signature (which is also essentially a zero knowledge proof of knowledge of a secret key) in the memo, which might have potentially interesting uses?

iamalwaysuncomfortable:
> I'm mildly in favour of keeping the full 64 bytes as unstructured data, since that would allow us to do things like store a signature (which is also essentially a zero knowledge proof of knowledge of a secret key) in the memo, which might have potentially interesting uses?

%100 agree. If there's future need for more structure, we can add other burn memo types

text/0039-burn-redemption-memo.md
iamalwaysuncomfortable: This MCIP has my approval either way. If other people think we should reserve some address space, that's fine, otherwise I think we're good to reserve space only as needed

iamalwaysuncomfortable (approved):

garbageslam (approved):
LGTM

PR #40, opened 2022-05-17
Token Metadata

cbeck88:
A proposal to support dynamically discoverable token metadata

[Rendered Proposal](https://github.com/garbageslam/mcips/blob/token-metadata/text/0040-token-metadata.md)

### DIFF
--- a/text/0040-token-metadata.md
+++ b/text/0040-token-metadata.md
@@ -1,0 +1,167 @@
+- Feature Name: `token_metadata`
+- Start Date: May 16 2022
+- MCIP PR: [mobilecoinfoundation/mcips#0040](https://github.com/mobilecoinfoundation/mcips/pull/0040)
+- Tracking Issue: (none yet)
+
+# Summary
+[summary]: #summary
+
+Specify a source of truth for metadata about new tokens, and a way for clients to discover this,
+by querying the consensus nodes.
+
+# Motivation
+[motivation]: #motivation
+
+[MCIP #25](https://github.com/mobilecoinfoundation/mcips/pull/0025) and [MCIP #37](https://github.com/mobilecoinfoundation/mcips/pull/0037)
+have specified ways for new tokens to exist on chain. However, for user-facing applications to make effective use of this feature,
+they need more information about a token than just a token id -- they ultimately need to know how they should present balance information
+to the user.
+
+One way to do this is to hard-code all metadata into clients, which might work initially. After this however, this will mean that new tokens
+that are created on chain can't actually be used by users until a client update has been released and the users have updated.
+
+(In the [MCIP #31](https://github.com/mobilecoinfoundation/mcips/pull/0025) implementation PR, we did add a proto enum called `KnownTokenIds`
+which can be shipped with clients, as a stop-gap measure.)
+
+If the clients can dynamically query this information, then a client update is not required for clients to make use of new tokens which are
+created, which significantly reduces the amount of time it takes for a new token to be available to users.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Token metadata includes the following:
+
+* Short (exchange ticker) name of a token, e.g. MOB. (This is not expected to be translated.)
+* Long (english) name of a token, e.g. MobileCoin. (This is expected to be translated, but the client must maintain the translation list.)
+* Display conversion factor, e.g. 10^12. This conversion factor converts from the smallest representable units (`u64`) which appear in the blockchain,
+  to the user-displayable unit. (Note that this is possibly not a power of 10.) `u64` amounts
+  which are obtained from `TxOut`'s on chain should be divided by this number, for the token id, before being displayed to the user.
+* (Optional) URL pointing to an SVG logo of the token. (This is expected to be scalable and appropriate for display on devices of many form factors.)
+* (Optional) URL pointing to an MCIP which proposed the token.
+
+The consensus network will provide an endpoint at which this metadata can be requested for one or more token ids.
+
+This data will be provided at startup time via a file to the consensus node. (In the future, we may consider to store this data on chain.)
+The node should require this data for any token which is configured to exist (has a minimum fee in `tokens.toml`).
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A token metadata proto has these fields
+
+```
+message TokenMetadata {
+    // The short name is a standard ticker symbol for the token
+    String short_name = 1;
+    // The long name is the standard english name for the token
+    String long_name = 2;
+    // The conversion factor is an unsigned 64 bit number expressed as a decimal string
+    String display_conversion_factor = 3;
+    // If present, a url pointing to a logo for the token
+    String logo_url = 4;
+    // If present, a url pointing to an MCIP which proposed the token
+    String mcip_url = 5;
+}
+```
+
+A consensus network client-facing API will be exposed to query for these objects given one or more token ids.
+
+The `tokens.toml` file schema is extended to contain these fields per token. The minting trust root should sign
+these fields.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This proposal
+
+* Adds to the complexity of configuring a network
+* Does not specify a mechanism that forces different nodes to have matching values
+  (we could choose to hash this data into responder id for instance)
+* Does not commit this data to the blockchain
+* Does not prevent the metadata from being changed -- only trust in the foundation not to sign another value
+* URL fields cannot be validated by users. For instance, we could include hash of the logo svg, and then they
+  could confirm the validity of the file they download.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The main rationale to justify the proposal is to try to make incremental progress increasing the discoverability
+of this important metadata.
+
+In the status quo, where different clients would have to hard-code this data, and then release it according to their
+release schedule, this may lead to a negative user experience. For example, if one user's wallet supports a new token id,
+and they decide to send it to another wallet of theirs that hasn't updated, or hasn't made a release that supports it yet,
+the new wallet will likely decide not to show this token to the user since it can't discover the metadata. From the user's
+point of view, their tokens have disappeared and they don't know how to get them back, so this is a negative user experience.
+
+With this proposal, all clients would be able to display all tokens with or without an update. So this seems to avoid
+a significant negative user experience.
+
+There are a number of ways this proposal could be strengthened however, described in the drawbacks section, around
+ensuring the integrity of this data.
+
+The main reasons not to do any of these is if we have uncertainty around whether / how this data should be prevented
+from being changed. For instance, it seems quite bad if the display conversion factor is changed after a token is launched.
+It might be desirable somehow for the logo or long name to be adjusted though.
+
+We could also allow this data to be committed to the chain, and for desired parts of it to be updateable, as part of a `MintConfigTx`
+as specified in MCIP #37. This would be a ledger format change, and so would require an enclave upgrade, increasing the amount of
+time it would take us to deliver such a change (and to not mitigate the negative user experience scenario described).
+
+Thus, the main rationale behind this form of the proposal would be as a stopgap that we can implement it quickly without an enclave change,
+and then provide an API to clients that improves a user experience, and makes it faster for clients to support new tokens.
+In a later proposal we could nail down the details of whether this data lives on or off chain, the ledger format change details if any,
+the consistency enforcement mechanism, the verifiability of the data by the end user, and so on.
+
+# Prior art
+[prior-art]: #prior-art
+
+In the Ethereum ecosystem, most tokens (fungible and non-fungible) have logos, but they are typically stored on centralized servers,
+and referred to via URLs.
+
+In [EIS-2569](https://eips.ethereum.org/EIPS/eip-2569) it was proposed to actually store SVG data on chain.
+
+Typically, wallets like metamask use services like Infura to access names and logos of tokens. There is not a standard way outside of these
+services to get this data for an arbitrary ERC-20.
+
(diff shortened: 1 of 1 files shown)


text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+
+* Short (exchange ticker) name of a token, e.g. MOB. (This is not expected to be translated.)
+* Long (english) name of a token, e.g. MobileCoin. (This is expected to be translated, but the client must maintain the translation list.)
+* Display conversion factor, e.g. 10^12. This conversion factor converts from the smallest representable units (`u64`) which appear in the blockchain,
+  to the user-displayable unit. (Note that this is possibly not a power of 10.) `u64` amounts
+  which are obtained from `TxOut`'s on chain should be divided by this number, for the token id, before being displayed to the user.
+* (Optional) URL pointing to an SVG logo of the token. (This is expected to be scalable and appropriate for display on devices of many form factors.)
+* (Optional) URL pointing to an MCIP which proposed the token.
+
+The consensus network will provide an endpoint at which this metadata can be requested for one or more token ids.
```
eranrund, line 42: Wanted to note that consensus already provides some information about configured tokens - https://github.com/mobilecoinfoundation/mobilecoin/blob/master/consensus/api/proto/consensus_config.proto#L33
so this proto file could be a good place for this to live. Probably still worth to provide a separate endpoint that only includes whats proposed in this PR since the config endpoint returns a lot of data clients don't necessarily care about.

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+
+This data will be provided at startup time via a file to the consensus node. (In the future, we may consider to store this data on chain.)
```
eranrund, line 44: I would suggest being explicit about this going into the `tokens.toml` file - I don't see a reason to have it be in a different file.

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A token metadata proto has these fields
+
+```
+message TokenMetadata {
+    // The short name is a standard ticker symbol for the token
+    String short_name = 1;
```
eranrund, line 55: Nit: I believe protobuf types are case sensitive so this should be `string`.

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+    // The long name is the standard english name for the token
+    String long_name = 2;
+    // The conversion factor is an unsigned 64 bit number expressed as a decimal string
+    String display_conversion_factor = 3;
```
eranrund, line 59:
```suggestion
    uint64 display_conversion_factor = 3;
```

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+The node should require this data for any token which is configured to exist (has a minimum fee in `tokens.toml`).
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A token metadata proto has these fields
+
+```
+message TokenMetadata {
+    // The short name is a standard ticker symbol for the token
```
eranrund, line 54: Might be nice to also include the token id here

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+    // If present, a url pointing to a logo for the token
+    String logo_url = 4;
+    // If present, a url pointing to an MCIP which proposed the token
+    String mcip_url = 5;
+}
+```
+
+A consensus network client-facing API will be exposed to query for these objects given one or more token ids.
+
+The `tokens.toml` file schema is extended to contain these fields per token. The minting trust root should sign
```
eranrund, line 69: Right now the `minting trust root` signs the governors and not the minimum fee, the only two configurable options per token. Maybe we should change the configuration to have two distinct sections - signed and unsigned, to make it clear what is under the signature and what isn't.
We can then add the signature to the responder id to ensure all signed configuration on all nodes is identical, which addresses one of the drawbacks mentioned below.

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+these fields.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This proposal
+
+* Adds to the complexity of configuring a network
+* Does not specify a mechanism that forces different nodes to have matching values
```
eranrund, line 78: I think we should just hash the signature of everything that is signed by the minting trust root.

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+  (we could choose to hash this data into responder id for instance)
+* Does not commit this data to the blockchain
+* Does not prevent the metadata from being changed -- only trust in the foundation not to sign another value
```
eranrund, line 81: the node operators would still need to accept the new file so the power is not fully in the foundation's hands

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+a significant negative user experience.
+
+There are a number of ways this proposal could be strengthened however, described in the drawbacks section, around
+ensuring the integrity of this data.
+
+The main reasons not to do any of these is if we have uncertainty around whether / how this data should be prevented
+from being changed. For instance, it seems quite bad if the display conversion factor is changed after a token is launched.
+It might be desirable somehow for the logo or long name to be adjusted though.
+
+We could also allow this data to be committed to the chain, and for desired parts of it to be updateable, as part of a `MintConfigTx`
```
eranrund, line 107: FWIW I don't think it's particularly hard to add this extra data to the mint config tx, especially at this point in time where the new enclave has not yet been signed. It is also backwards-compatible since adding new fields would not break the proto.
The one disadvantage of adding it to the `MintConfigTx` is that it is more stuff to get correctly when issuing this transaction, a typo could very easily result in a lot of confusion to users. And realistically IMO the metadata is very unlikely to be changed over time.

text/0040-token-metadata.md
garbageslam: okay, makes sense

also it's worth discussing, maybe this should be just "number of decimals" since that's what everyone seems to do, when I was researching for prior art

text/0040-token-metadata.md
garbageslam: maybe it's okay for the governers to sign this instead of the minting trust root?

eranrund: In which case it will likely need to go `MintConfigTx`... or it will be a pretty convoluted process of back and forth between the governors and then the node operators to update the new tokens.toml

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+
+* Adds to the complexity of configuring a network
+* Does not specify a mechanism that forces different nodes to have matching values
+  (we could choose to hash this data into responder id for instance)
+* Does not commit this data to the blockchain
+* Does not prevent the metadata from being changed -- only trust in the foundation not to sign another value
+* URL fields cannot be validated by users. For instance, we could include hash of the logo svg, and then they
+  could confirm the validity of the file they download.
+
+# Rationale and alternatives
```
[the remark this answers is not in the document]
garbageslam: that seems pretty reasonable!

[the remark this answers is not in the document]
garbageslam: I think that's simpler than this proposal and it meets the goal of being a stopgap for clients, described in the Rationale. We can think about what a more permanent solution looks like after that (thanks for your thoughts on that also!)

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+>    Decimals (required)
+>    DefaultFrozen (required)
+>    URL (optional)
+>    MetaDataHash (optional)
+
+In Cardano, native assets can be created before being [registered with the token registry](https://developers.cardano.org/docs/native-tokens/token-registry/How-to-prepare-an-entry-for-the-registry-NA-policy-script).
+
+Only a name is required to create the asset. When subsequently registering an asset, a "description" is required. Then a ticker, url, logo, and decimals are all considered optional.
+
+In the Cardano registry, metadata can be deleted and updated, but it requires a signature from the creator.
```
remoun, line 144: Nit: These 3 sentences are related, so let's group them into a paragraph

```suggestion
In Cardano, native assets can be created before being [registered with the token registry](https://developers.cardano.org/docs/native-tokens/token-registry/How-to-prepare-an-entry-for-the-registry-NA-policy-script). Only a name is required to create the asset. When subsequently registering an asset, a "description" is required. Then a ticker, url, logo, and decimals are all considered optional. In the Cardano registry, metadata can be deleted and updated, but it requires a signature from the creator.
```

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+
+Depending on the outcome of discussion, we could increase the scope of this proposal significantly, or scope it more narrowly and make a follow-up MCIP.
+
+We anticipate that we may want more metadata items than this as well, or to change the names of them.
+
+It is also possible that we may want to handle internationalization differently, either requiring that there is a source of truth for how to translate token names
+into other languages, or not providing a long name for a token at all, and leaving that entirely to the clients.
+
+We suspect that in the end we will want all of this metadata to go onto the chain, and the main barrier to that is implementation complexity. But this may be considerably
+simpler now that we have `MCIP 37` `MintConfigTx`'s.
```
remoun, line 167:
```suggestion
simpler now that we have `MintConfigTx`'s (MCIP #37).
```

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
+- Feature Name: `token_metadata`
+- Start Date: May 16 2022
+- MCIP PR: [mobilecoinfoundation/mcips#0040](https://github.com/mobilecoinfoundation/mcips/pull/0040)
+- Tracking Issue: (none yet)
```
remoun, line 4: Let's file a tracking issue :)

text/0040-token-metadata.md
```diff
@@ -0,0 +1,167 @@
+- Feature Name: `token_metadata`
+- Start Date: May 16 2022
+- MCIP PR: [mobilecoinfoundation/mcips#0040](https://github.com/mobilecoinfoundation/mcips/pull/0040)
```
remoun, line 3: The leading zeros unnecessary and potentially confusing IMO.

```suggestion
- MCIP PR: [mobilecoinfoundation/mcips#40](https://github.com/mobilecoinfoundation/mcips/pull/40)
```

text/0040-token-metadata.md
```diff
@@ -0,0 +1,163 @@
...
+created, which significantly reduces the amount of time it takes for a new token to be available to users.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Token metadata includes the following:
+
+* Short (exchange ticker) name of a token, e.g. MOB. (This is not expected to be translated.)
+* Long (english) name of a token, e.g. MobileCoin. (This is expected to be translated, but the client must maintain the translation list.)
+* Display conversion factor, e.g. 10^12. This conversion factor converts from the smallest representable units (`u64`) which appear in the blockchain,
```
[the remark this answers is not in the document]
garbageslam: i think for bitcoin, the fundamental unit is satoshi?

i think in algorand standard assets, they have a "unit name" field, so that could be e.g. picoMob, satoshi, etc. for us?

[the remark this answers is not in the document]
garbageslam: i am not sure how SI prefix stuff interacts with translations, so like, i think usually you could not translate `pico`, then translate `Mob`, then concatenate them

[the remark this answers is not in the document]
the-real-adammork: https://www.metricmethods.com/Resources.php

[Chinese](https://www.metricmethods.com/SI_chinese.html)	[French](https://www.metricmethods.com/SI_french.html)	[German](https://www.metricmethods.com/SI_german.html)	[Greek](https://www.metricmethods.com/SI_greek.html)
[Italian](https://www.metricmethods.com/SI_italian.html)	[Japanese](https://www.metricmethods.com/SI_japanese.html)	[Russian](https://www.metricmethods.com/SI_russian.html)	[Spanish](https://www.metricmethods.com/SI_spanish.html)

Im ok with unit_name, but that could end up being less translatable than an SI prefix.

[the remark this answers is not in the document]
the-real-adammork: thinking more about it, probably best to choose the most flexible option which would be a `unit_name` field

text/0040-token-metadata.md
```diff
@@ -0,0 +1,163 @@
...
+
+```
+message TokenMetadata {
+    // The short name is a standard ticker symbol for the token
+    String short_name = 1;
+    // The long name is the standard english name for the token
+    String long_name = 2;
+    // The conversion factor is an unsigned 64 bit number expressed as a decimal string
+    String display_conversion_factor = 3;
+    // If present, a url pointing to a logo for the token
```
the-real-adammork, line 60: add `unit_name`

```
    // The unit name is the name of the fundamental unit of this token
    String unit_name = ...;
```

cbeck88:
We merged MCIP 59, this one is somewhat mooted, but maybe we should leave it open and rewrite it later if we decide to do token metadata on chain.

nick-mobilecoin:
Adding Final Comment Period to propose Closing this.

[59](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0059-token-metadata-document.md) more or less does what this does.
[59](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0059-token-metadata-document.md)
Does say:
> This [59] proposal differs from an earlier proposal https://github.com/mobilecoinfoundation/mcips/pull/0040, in that it is a stop-gap mechanism which does not post metadata to the blockchain.

However it doesn't look like this actually handled posting to blockchain.

cbeck88:
this needs a lot of rework anyways if we want to do on-chain metadata later, let's just rewrite it at that time

PR #41, opened 2022-05-19
Gift code recoverable fee history

iamalwaysuncomfortable:
A proposal to introduce a recoverable fee history into FOG compatible gift codes.

[Rendered Proposal](https://github.com/iamalwaysuncomfortable/mcips/blob/main/text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md)

### DIFF
--- a/text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
+++ b/text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
@@ -27,13 +27,13 @@ gift code flow.
 # Motivation
 [motivation]: #motivation
 
-Normal recoverable transaction history isn't possible for gift codes because at the point when a gift code is funded , 
+Normal recoverable transaction history isn't possible for gift codes because at the point when a gift code is funded, 
 it is not yet known what account will redeem this gift code. When funding or cancelling a gift code, the sender may want
 to be able to have a recoverable history for the fees they paid to do so.
 
-Similarly, when a gift code is "redeemed" the amount sent to the change address will differ from the amount received in
-the protobuf message from the sender by an amount equal to the fee paid to redeem it. This may be confusing to gift code
-recipients if they are not able to account for the fee that they paid to redeem it.
+Similarly, when a gift code is "redeemed" the amount sent to the change address will differ from the amount of the TxOut
+the receiver is able to unblind with the data received from the sender by an amount equal to the fee paid to redeem it. 
+This may be confusing to gift code recipients if they are not able to account for the fee that they paid to redeem it.
 
 Thus we propose modifications (explained below) to the gift code memo types in order to create recoverable fee history.
 


text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
+- Feature Name: Recoverable-fee-history-For-FOG-compatible-gift-codes
+- Start Date: May 18
```
eranrund, line 2:
```suggestion
- Start Date: May 18 2022
```

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
...
+that allows the receiver to record up to 64 utf-8 bytes to document any desired information about the gift code.
+
+The memos which record this information are shown in the table below:
+
+| Memo type bytes | Name                                              |
+| -----------     | -----------                                       |
+| 0x0201          | Gift code funding memo                            |
+| 0x0202          | Gift code cancellation memo                       |
+| 0x0002          | Gift code sender memo                       |
+
```
eranrund, line 34: I don't think there's a need to repeat all this information here, it's available in MCIP 32.

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
...
+Similarly, when a gift code is "redeemed" the amount sent to the change address will differ from the amount received in
+the protobuf message by an amount equal to the fee paid to redeem it. This may be confusing to gift code recipients if 
+they are not able to account for the fee that they paid to redeem it.
+
+Thus we propose modifications to the gift code memo types in order to create recoverable fee history.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Each memo type will have allocate 7 bytes for an unsigned 56-bit number which represents the fee amount paid in each
```
eranrund, line 57: Should mention (and maybe link) the original RTH MCIP, that also uses 7 bytes for representing the fee.

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
...
+| ---------- | ---- |
+| 0 - 4      |  First 4 bytes of blake2b over `tx_out_public_key` of gift code tx out |
+| 4 - 64     | "Note": A UTF-8 string with that is either exactly 60 bytes or null terminated if under 60 bytes |
+
+### Proposed change
+| Byte range | Item |
+| ---------- | ---- |
+| 0 - 4      |  First 4 bytes of blake2b over `tx_out_public_key` of gift code tx out |
+| 4 - 11      | Big-endian bytes of fee amount, as an unsigned 56-bit number |
+| 11 - 64     | "Note": A UTF-8 string with that is either exactly 60 bytes or null terminated if under 60 bytes |
```
eranrund, line 79:
```suggestion
| 11 - 64     | "Note": A UTF-8 string with that is either exactly 53 bytes or null terminated if under 53 bytes |
```

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
...
+### Current description
+| Byte range | Item |
+| ---------- | ---- |
+| 0 - 64     | "Note": A UTF-8 string with that is either exactly 64 bytes or null terminated if under 64 bytes |
+
+### Proposed change
+| Byte range | Item |
+| ---------- | ---- |
+| 0 - 7      | Big-endian bytes of fee amount, as an unsigned 56-bit number |
+| 7 - 64     | "Note": A UTF-8 string with that is either exactly 64 bytes or null terminated if under 64 bytes |
```
eranrund, line 107:
```suggestion
| 7 - 64     | "Note": A UTF-8 string with that is either exactly 57 bytes or null terminated if under 57 bytes |
```

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
...
+The rationale behind this is to create consistent accounting.
+
+However it could potentially be said that this is not strictly necessary and that it should be left up to implementing 
+apps to decide what to do with the utf-8 bytes. In a regime limited to 64 bytes or less, 7 bytes could be useful. 
+Therefore alternatives could be not to implement fee tracking at all OR to only use 6 bytes to track the fee.
+
+# Prior art
+[prior-art]: #prior-art
+
+[MCIP #32 - FOG-compatible-gift-codes](https://github.com/mobilecoinfoundation/mcips/blob/main/text/0032-fog-compatible-gift-codes.md)
```
eranrund, line 127: I think the original RTH MCIP that chose 7 bytes for the fees is also part of the prior art.

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+The merits of recoverable fee history over the utf-8 bytes they replace
```
eranrund, line 132: I don't think this is an unresolved question, if the MCIP gets accepted then we want the fee bytes and if it doesn't (in its current form) then we don't...

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,137 @@
...
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Future extensions of the gift code memos or creation of new gift code memo types
```
eranrund, line 137: I don't think there's much to expand on in this particular MCIP.

eranrund (changes requested):

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
iamalwaysuncomfortable: I saw it in the frame of the core topic of debate of the MCIP, but I'm agnostic about how to frame it, I can just put "None"

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
iamalwaysuncomfortable: Slimmed down, but I think it's worth keeping the table.

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,129 @@
...
+| 0x0002          | Gift code sender memo                       |
+
+These memos do not currently allocate any bytes for tracking transaction fees. This MCIP proposes to allocate 7 bytes to
+each memo representing an unsigned 56 bit number which can be used to track the fee amount paid in each stage of the 
+gift code flow.
+
+# Motivation
+[motivation]: #motivation
+
+Normal recoverable transaction history isn't possible for gift codes because at the point when a gift code is funded ,
```
eranrund, line 30:
```suggestion
Normal recoverable transaction history isn't possible for gift codes because at the point when a gift code is funded, 
```

text/0041-Recoverable-fee-history-for-FOG-compatible-gift-codes.md
```diff
@@ -0,0 +1,129 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Normal recoverable transaction history isn't possible for gift codes because at the point when a gift code is funded , 
+it is not yet known what account will redeem this gift code. When funding or cancelling a gift code, the sender may want
+to be able to have a recoverable history for the fees they paid to do so.
+
+Similarly, when a gift code is "redeemed" the amount sent to the change address will differ from the amount received in
+the protobuf message from the sender by an amount equal to the fee paid to redeem it. This may be confusing to gift code
```
eranrund, line 35: It would also differ from the amount in the giftcode `TxOut` itself

eranrund (approved):

jcape:
Couple offhand notes:
1. "FOG" is not an acronym, it doesn't need to be fully capitalized.
2. The name is super long, let's pick something shorter, e.g. `gift-code-fee-memos` or something
3. Please name the file using all lower case to match the existing stuff.
4. I'm dying to know how all of us looked at MCIP #3 and MCIP #4 and nobody asked why we're using _big endian_ numbers? Why are we doing this?

garbageslam (changes requested):
I think it is a good idea to actually update the text of the earlier MCIPs?

It will make it easier for people to read if it just has the final state, git will record the initial state if it's interesting for historical purposes

iamalwaysuncomfortable:
> I think it is a good idea to actually update the text of the earlier MCIPs?

Sure, I'm ambivalent. People directed me to make a fixer MCIP  to keep in line with past process (instead of making a direct commit to MCIP 32 or having PRs only be mcips) & to also give people a chance to comment, so I did that.

I'm ambivalent. Once people reach consensus here on whether it should be a fixer MCIP vs. changes to the existing MCIP, we'll go whatever route consensus here decides.

iamalwaysuncomfortable:
> why we're using big endian numbers

This is a good point. It was kept consistent with the previous MCIPs but little endian would've been the better choice I think.

garbageslam:
> This is a good point. It was kept consistent with the previous MCIPs but little endian would've been the better choice I think.

It is simpler to be consistent everywhere in the memos.

I think the only reason was Koe had proposed that we use 7-byte numbers instead of 8-byte for the fee, to cram more data in there. It's easier to figure out what a 7-byte big endian number looks like than a 7-byte little endian number.

iamalwaysuncomfortable:
> It's easier to figure out what a 7-byte big endian number looks like than a 7-byte little endian number.

Also a good point

jcape:
Closing in favor of #32 changes.

PR #43, opened 2022-06-09
Block Metadata

remoun:
A proposal to add block metadata containing the quorum set and attested verification report of the consensus node that externalized that block.

[Rendered proposal](https://github.com/mobilecoinfoundation/mcips/blob/block-metadata/text/0043-block-metadata.md)


text/0043-block-metadata.md
```diff
@@ -0,0 +1,193 @@
...
+with historical consensus enclave AVRs, up to when consensus nodes publish
+`BlockMetadata` with their AVRs.
+
+The lookup data will be in the following TOML format:
+```toml
+[[node]]
+responder_id = ""  # Responder ID for the consensus node.
+avr = ""  # AVR for the node's enclave, includes pubkey.
+first_block_index = X  # Index of the first block signed with this enclave.
+last_block_index = Y  # Optional index of the last block signed with this enclave.
```
iamalwaysuncomfortable, line 141: The AVRs change with changes in message signing key (AVRs capture a unique signature of all the data and code and thus changes if the message signing pubkey changes).

```suggestion
responder_id = ""  # Responder ID for the consensus node.
avr = ""  # AVR for the node's enclave, includes pubkey.
message_signing_key = # Message signing key the AVR was associated with
first_block_index = X  # Index of the first block signed with this enclave.
last_block_index = Y  # Index of the last block signed with this enclave.
```

iamalwaysuncomfortable: Also to note.
1. There's missing data for each node_uri/signing key pair
2. TOML doesn't have a great null option, it's better perhaps if we don't include the fields in the bootstrap file

nick-mobilecoin, line 141: Since these are historical, we should have a cut over time.  Should `last_block_index` be required here?

nick-mobilecoin (approved):

text/0043-block-metadata.md
remoun: > Since these are historical, we should have a cut over time. Should `last_block_index` be required here?

Good catch!

```suggestion
last_block_index = Y  # Index of the last block signed with this enclave.
```

> It probably wouldn't be strictly required, but it'd make bootstrapping dbs with it more of a pain/error prone since you'd have to somehow access the next record and be sure that's actually the next record.

Not necessarily. We just need to verify that the current block's index is within the range.

remoun: What would we gain from including the pub key? Since the AVR includes the pub key, including the pub key in this config seems redundant.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,193 @@
...
+
+    /// Block signature.
+    BlockSignature signature = 3;
+
+    /// Additional signed metadata about this block.
+    BlockMetadata metadata = 4;
+}
+```
+
+Since this is a protobuf-additive change, we don't need a new `ArchiveBlock`
```
wjuan-mob, line 98: Would it make sense to have a blockversion field here to know if we expect the new archiveblock format with the added metadata vs the old format without it?

text/0043-block-metadata.md
```diff
@@ -0,0 +1,193 @@
...
+TOML format:
+```toml
+[[node]]
+message_signing_pub_key = ""  # Pub key, as PEM-encoded data or a file path containing such data.
+first_block_index = X  # Index of the first block signed with this key.
+last_block_index = Y  # Optional index of the last block signed with this key.
+```
+
+## Historical AVRs
+With `BlockMetadata`, consumers of the blockchain no longer need a Watcher DB.
```
eranrund, line 127: Its worth noting that the watcher is currently used to provide block timestamps, so I think consumers of the blockchain that need that information will still need it. The root of the issue is that timestamps are optional, and only available for blocks signed by an enclave. When syncing blocks from S3 (e.g. when bootstrapping a new node, or when a node does catchup) I believe this results in it not having timestamps for those blocks.
The current workaround is having the watcher sync blocks from all nodes, and that guarantees that once all blocks have been synced from all nodes for a given block index, at least one of them will have a proper signature/timestamp.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,193 @@
...
+[[node]]
+responder_id = ""  # Responder ID for the consensus node.
+avr = ""  # AVR for the node's enclave, includes pubkey.
+first_block_index = X  # Index of the first block signed with this enclave.
+last_block_index = Y  # Index of the last block signed with this enclave.
+```
+
+## Block Version bump
+Block version 3 (release v1.3) will make the `block_metadata` required, at which
+point consensus nodes and clients should reject blocks without metadata.
```
eranrund, line 146: How does work when a node is performing catch up (for example when syncing blocks that were formed when it was down). Does it arbitrarily picks up metadata from one of the sources it syncs from?

text/0043-block-metadata.md
remoun: It's already included, under `ArchiveBlock.v1.block.version`

text/0043-block-metadata.md
remoun: The `ArchiveBlocks` in S3 already include the `BlockSignature` from which watcher derives the block timestamp. AFAICT that is sufficient for catching up.

eranrund: Unless the code changed, the [ledger sync code](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/ledger/sync/src/ledger_sync/ledger_sync_service.rs#L211) does not append the signature. It doesn't do that since the node has no way of validating that this signature actually came from a legitimate node ( in theory someone could replace the s3 contents with whatever they want right now). We made the decision that a node's local ledger only contains signature generated by the node itself, which also means you do not end up with a ledger that has a mishmash of signatures from multiple nodes.

remoun: The AVR + signing key validation is integral for ensuring the metadata comes from a trusted consensus node, and since the metadata includes the AVR, we can validate the enclave key used for the block signature, and thus authenticate the signature.

> Unless the code changed, the [ledger sync code](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/ledger/sync/src/ledger_sync/ledger_sync_service.rs#L211) does not append the signature.

I switched to adding the whole `BlockData` in https://github.com/mobilecoinfoundation/mobilecoin/pull/2052; it sounds like that should not be merged until the AVR + signing key validation is.

eranrund: That will only work going forward.

Right now it is possible there are blocks in S3 what contain no signature (e.g. if a node performed catchup at any point, it will then upload signature-less blocks into its S3 bucket).
Looking at https://github.com/mobilecoinfoundation/mobilecoin/blob/block-metadata-in-ledger/ledger/sync/src/ledger_sync/ledger_sync_service.rs, unless I am missing something, I am not seeing a path that prioritizes blocks that do we a signature and timestamp.

This MCIP is not necessarily the right place to continue this discussion, but my point is we should be careful about hinting the watcher can be removed. Unless the ledger syncing code changes to download a block from multiple sources until a version of it with a signature/timestamp is located, the watcher is still needed.

remoun: I see. Thank you for flagging!

text/0043-block-metadata.md
iamalwaysuncomfortable: I believe it would make the toml more clear to read since the avr is pretty verbose, but it's not specifically necessary.

text/0043-block-metadata.md
iamalwaysuncomfortable: Tracing the code - I believe the timestamp gets added within the logic in this call: https://github.com/mobilecoinfoundation/mobilecoin/blob/master/watcher/src/verification_reports_collector.rs#L92 - would that be correct?

And then it checks the signer key every poll interval: https://github.com/mobilecoinfoundation/mobilecoin/blob/master/watcher/src/verification_reports_collector.rs#L220

At the moment we're working on a bootstrap file for this to provide the timestamps historically, but any new client that "replaces watcher" would have to include this attest() call to add the timestamp and explicitly check for AVRs I think. 

So it seems like we'll maybe have to wait until we have a block streaming component actively collecting these AVRs.

remoun: Yeah, I've arrived at a similar conclusion. I've updated the wording to "consumers of the blockchain may not need a Watcher DB."

text/0043-block-metadata.md
```diff
@@ -0,0 +1,193 @@
...
+## Signing key history
+MobileCoin Foundation will publish a mapping/list of node_id to a range of block
+indexes, which node operators and consumers can use to verify the metadata
+signatures.
+
+This mapping will contain multiple node_id:block_range entries, in the following
+TOML format:
+```toml
+[[node]]
+message_signing_pub_key = ""  # Pub key, as PEM-encoded data or a file path containing such data.
```
eranrund, line 121: The matching PR currently uses hex bytes for this, but I do think PEM is the right way to go.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,193 @@
...
+last_block_index = Y  # Index of the last block signed with this enclave.
+```
+
+## Block Version bump
+Block version 3 (release v1.3) will make the `block_metadata` required, at which
+point consensus nodes and clients should reject blocks without metadata.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
eranrund, line 152: I think another drawback worth mentioning is that `mobilecoind` and `full-service` users would now need to get this historical AVR file if they want to have proper validation. If we make it optional to maintain backwards compatibility then many might skip doing that.

eranrund (approved):

text/0043-block-metadata.md
remoun: Agreed, added support for inline PEM as well as PEM files or hex-encoded keys.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,238 @@
...
+# Optional index of the last block signed with this enclave.
+last_block_index = Y
+[node.avr]
+# Hex-encoded signature bytes
+sig = "..."
+# An array of hex-encoded pub key bytes
+chain = ["x", "y"]
+# IAS HTTP response body
+http_body = '{...}'
+```
```
jcape, line 168: Is there an advantage to doing this vs. `avr_sig avr_chain`, `avr_http_body`?

remoun: It maintains the grouping of related fields by type rather than a convention of common prefix, and keeps the JSON and TOML representations compatible.

With that said, I'm open to changing to `avr_*` if that proves better.

text/0043-block-metadata.md
remoun: To close the loop on this, https://github.com/mobilecoinfoundation/mobilecoin/pull/2216 handles catch-up mode

text/0043-block-metadata.md
```diff
@@ -0,0 +1,238 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+This is a long-awaited feature for the MobileCoin blockchain.
+
+The AVR proves that the block was signed by an enclave, as the pubkey used to
+sign the block is in the AVR. Signing the metadata with the SCP message signing
+key proves that the block and metadata were generated by the holder of that
+private key.
```
iamalwaysuncomfortable, line 37: This is more of a summary than a motivating statement

```suggestion
Currently if one wants to prove blocks were generated by the MobileCoin network (say
instead of a forged chain) - the information required to prove this is hard to obtain.

In order to ensure chain integrity to participants in the mobilecoin network
and those auditing it. The following information is proposed to be distributed
on the blockchain itself:
1. An AVR that proves a block was signed by an SGX enclave signed by the Intel chain
of trust
2. Block Metadata including the node's quorum set, AVR, and SCP message signing key 
signed by the SCP message signing key to prove that the node was the holder of that
private key when the block was signed
```

text/0043-block-metadata.md
```diff
@@ -0,0 +1,238 @@
...
+  ]
+}
+```
+
+## Historical AVRs
+With `BlockMetadata`, consumers of the blockchain may not need a Watcher DB.
+Ledger DB can default to reading AVRs from `BlockMetadata` (using the
+aforementioned signing key history to authenticate the metadata signature),
+using the AVR to verify the enclave key used for the block signature, and fall back
+to a lookup table containing AVRs for older blocks.
```
iamalwaysuncomfortable, line 146:
```suggestion
## Historical AVRs
The Block Signing Key used to sign the transaction data within a block is included within the
`isvQuoteBody` field of an Intel AVR. By having AVRs available on chain, anyone can verify
the reported block signing key is correct and that the block was signed by a valid enclave.
Thus With `BlockMetadata`, consumers of the blockchain may not need a Watcher DB
because Ledger DB can default to reading AVRs from `BlockMetadata` 
```

text/0043-block-metadata.md
```diff
@@ -0,0 +1,238 @@
...
+        "http_body": "{...}"
+      }
+    }
+  ]
+}
+```
+
+## Block Version bump
+Block version 3 (release v1.3) will make the `block_metadata` required, at which
+point consensus nodes and clients should reject blocks without metadata.
```
iamalwaysuncomfortable, line 190:
```suggestion
Block version 3 (release v1.3) will make the `block_metadata` required, at which
point consensus nodes and clients should reject blocks without metadata.

No later than release 1.4. Utilities for verifying this block metadata will be
distributed as well for network participants and auditors to use.
```

text/0043-block-metadata.md
```diff
@@ -0,0 +1,238 @@
...
+#### AVR History TOML format
+```toml
+[[node]]
+# Responder ID for the consensus node.
+responder_id = ""
+# Index of the first block signed with this enclave.
+first_block_index = X
+# Optional index of the last block signed with this enclave.
+last_block_index = Y
+[node.avr]
```
iamalwaysuncomfortable, line 161: Given some nodes may not actually have Block Signing Keys recorded within historical blocks OR may have signing keys that have no corresponding AVRs (because of watcher outages, etc.), we may consider putting these within the bootstrap to indicate when in the chain history these null states existed.

```suggestion
# Record block signing key for the node
block_signing_key = ""
[node.avr]
```

iamalwaysuncomfortable (commented):
Some suggestions for greater clarity + reinforcement that the bootstrap format suggested is final (modulo the block signing key) + questions as to whether we need to add the quorum set to our historical records.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,238 @@
...
+Ledger DB can default to reading AVRs from `BlockMetadata` (using the
+aforementioned signing key history to authenticate the metadata signature),
+using the AVR to verify the enclave key used for the block signature, and fall back
+to a lookup table containing AVRs for older blocks.
+
+MobileCoin Foundation will publish a bootstrap file defining this lookup table
+with historical consensus enclave AVRs, up to when consensus nodes publish
+`BlockMetadata` with their AVRs.
+
+#### AVR History TOML format
```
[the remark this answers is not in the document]
remoun: It could be nice, but I don't know of any historical data containing quorum set info.

text/0043-block-metadata.md
iamalwaysuncomfortable: It's true that the AVRs do include them, however there are block ranges in which Block Signing Keys weren't included on chain (I believe previous catchup mode invocations simply avoided including it, so any ranges a consensus node had to play catchup are devoid of a signing key I think) - and we may want to indicate ranges that those Block Signing Keys were not present for a given (Range, ResponderID) tuple so that we don't automatically fail the blocks.

iamalwaysuncomfortable: This wasn't immediately clear at the start of the project and only became clear afterward

remoun:
Extending FCP to Tuesday, given the long weekend.

remoun:
Sorry, I meant Wednesday.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+## Signing key history
+MobileCoin Foundation will publish a mapping/list of node_id to a range of block
+indexes, which node operators and consumers can use to verify the metadata
+signatures.
+
+This mapping will contain multiple node_id:block_range entries, in the formats defined below.
+
+#### Key History TOML format
+```toml
+[[node]]
```
eranrund, line 125: It might be worth including the responder id here - at the very least it would make it easier for humans to figure out where each key is supposed to be coming from.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Historical AVRs
+The Block Signing Key used to sign the transaction data within a block is
+included within the `isvQuoteBody` field of an Intel AVR. By having AVRs
+available on chain, anyone can verify the reported block signing key is correct
+and that the block was signed by a consensus node with a valid enclave and a
+recognized SCP message signing key.
+
+Thus With `BlockMetadata`, consumers of the blockchain may not need a Watcher DB
+because Ledger DB can default to reading AVRs from `BlockMetadata`, , and fall
```
eranrund, line 155:
```suggestion
because Ledger DB can default to reading AVRs from `BlockMetadata`, and fall
```

text/0043-block-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+## Historical AVRs
+The Block Signing Key used to sign the transaction data within a block is
+included within the `isvQuoteBody` field of an Intel AVR. By having AVRs
+available on chain, anyone can verify the reported block signing key is correct
+and that the block was signed by a consensus node with a valid enclave and a
+recognized SCP message signing key.
+
+Thus With `BlockMetadata`, consumers of the blockchain may not need a Watcher DB
```
eranrund, line 154: I believe the main purpose of the watcher these days other than keeping track of history is to provide guaranteed timestamps. So consumes needing block timestamps would still need it, since timestamps are coupled to block signatures and those are only available on blocks originated from an enclave and not via catch up

eranrund (changes requested):
Some small nits and a suggestion to include the responder id in the key history files.

text/0043-block-metadata.md
remoun: I hesitate to add a field we cannot use (because there's no way to get a responder ID from any block data); humans can include the responder ID in a comment.

eranrund: Whats the difference between having it here and having it in the `AVR History` file?

text/0043-block-metadata.md
remoun: Added, and moved this note about removing watcher to Future Possibilities, since it felt increasingly out of place here.

text/0043-block-metadata.md
```diff
@@ -0,0 +1,359 @@
...
+1. Enables block streaming and other consumers to include and validate the AVRs
+   for persisted blocks, which opens a path to replacing mobilecoind and watcher
+   in Fog.
+
+   Note that block timestamps come from the block signature, and nodes that are
+   catching up to the latest consensus (including mobilecoind, full-service, and
+   consensus nodes that are catching up) do not append any block signature,
+   relying on a Watcher to update the signatures. Replacing Watcher thus entails
+   updating this catch-up mode to handle block signatures as well as metadata,
+   perhaps reusing SCP.
```
eranrund, line 359:
```suggestion
   relying on a Watcher to provide timestamps (by syncing all blocks from multiple sources,
   at least one is guaranteed to contain a signature and a timestamp).
   Replacing Watcher thus entails updating this catch-up mode to handle block signatures as well
   as metadata, perhaps reusing SCP quorum sets.
```

(we don't store SCP messages so that is not reusable, but perhaps a node can choose to accept a signature(+timestamp) if a quorum set check can be satisfied after having fetched the same block from multiple sources. Either way that is out of scope for this MCIP)

eranrund (approved):
Hey @remoun, I am still wondering about including the responder id in the file: https://github.com/mobilecoinfoundation/mcips/pull/43/files/b84c80154548e6fd9a356ce9b42b8e02677c1a4b#diff-fc4c129a2799b7945224cb424bdde32a59a0c816648af8ec6f3db7c97ff1ca3f but I don't think thats a blocker.

remoun:
PTAL. I discussed the responder_id addition with James, and we agreed to add it to the metadata (contents) as well as the 2 proposed configs.

I'm extending FCP till tomorrow; hoping to merge this Fri AM PDT.

eranrund (approved):

text/0043-block-metadata.md
remoun: IIUC we would then include these ranges in the file with no AVR, to signify that this is a range known to lack that data.

PR #45, opened 2022-07-15
Reserved Token IDs

jcape:
[Rendered Proposal](https://github.com/jcape/mcips/blob/reserved-token-ids/text/0045-reserved-token-ids.md)

### DIFF
--- a/text/0045-reserved-token-ids.md
+++ b/text/0045-reserved-token-ids.md
@@ -1,6 +1,6 @@
 - Feature Name: `reserved_token_ids`
 - Start Date: 2022-07-14
-- MCIP PR: [mobilecoinfoundation/mcips#0000](https://github.com/mobilecoinfoundation/mcips/pull/0000)
+- MCIP PR: [mobilecoinfoundation/mcips#0045](https://github.com/mobilecoinfoundation/mcips/pull/45)
 - Tracking Issue: N/A
 
 # Summary
--- a/text/0045-reserved-token-ids.md
+++ b/text/0045-reserved-token-ids.md
@@ -11,7 +11,7 @@ MobileCoin 1.2 introduced the ability for the network to operate on different as
 # Motivation
 [motivation]: #motivation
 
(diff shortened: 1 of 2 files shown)


text/0045-reserved-token-ids.md
```diff
@@ -0,0 +1,68 @@
...
+
+We will need to allocate token IDs for development and testing purposes on our various networks. We also wish to ensure a given token ID is unique for an underlying asset across our networks: development and test network(s) (e.g. AlphaNet, TestNet, etc.), and our main network (i.e. MainNet).
+
+It is also critically important that a given token ID not be re-used within a network, because (presumably) the new token is not a direct successor token to the previously existing token, and it is not considered reasonable to expect the original token to be completely burned via transaction to the verifiable burn address. If a holder of a particular token loses their private key, then the contents of their account cannot be verifiably burned.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+As a result, the primary purpose of this change is to ensure that the same token ID is not re-used on a given network, but secondarily to ensure that a given token ID is also not re-used *between* networks. 
+
```
nick-mobilecoin, line 22: I would think that we would want to allow main net token ids on the test net in order to better simulate tokens on mainnet, prior to them being on mainnet.
Is the desire that one spins up their own local network for these kind of tests and sanity checks?

nick-mobilecoin (approved):

cbeck88 (changes requested):
I am worried that this compromises our ability to properly test clients in testnet before we go live in mainnet with important synthetic assets.

I think this proposal should be at a minimum be reworked to explain how clients are supposed to be tested if we adopt this proposal. I think this proposal is non-viable because of the difficulty it creates for this.

I think we should also have a more serious discussion of the alternatives in rationale and alternatives. There are easy ways to handle "de-confliction" of networks, which is an important problem that affects all features and not just confidential tokens. Instead of fragmenting configuration across all networks maybe we should come up with unified way to deal with de-confliction, and then everything else will become simpler.

garbageslam:
I posted a proposal here to explain what I have in mind: https://github.com/mobilecoinfoundation/mcips/pull/49

What I would like is that, token ids that appear in prod are okay to appear in test and dev, and token ids that appear in test are okay to appear in dev. This allows that we can roll out a specific new token id through testing at each layer of the system before we go live, and we don't need to have network-specific token metadata.

PR #47, opened 2022-07-15
Update 0043-block-metadata.md

colinlowenberg:
Fixing the format on the header/metadata from * to - as bullets. This formatting is required by the parser for our website's navigation menu.

The rendered proposal is now re-rendered on our website:
https://mobilecoin.com/proposals/block-metadata

Alternative solution: We could update our parser to support both types of bullets.


jcape (approved):

jcape:
Could you submit a PR to update the template as well?

garbageslam (approved):

PR #48, opened 2022-07-15
Update 0037-minting.md

colinlowenberg:
Adding a user-friendly name. This will look nice on our website here: https://mobilecoin.com/proposals/minting


jcape:
That is not what that header element is for.

PR #49, opened 2022-07-25
A proposal for chain-id's, to help avoid using the wrong network

garbageslam:
[Rendered Proposal](https://github.com/garbageslam/mcips/blob/network-ids/text/0049-chain-ids.md)


text/0049-network-ids.md
```diff
@@ -0,0 +1,183 @@
...
+* This has the benefit that the error messages when networks are mixed will be consistent and much clearer.
+
+Note that while we are partly motivated by reducing complexity and easing testing around MCIP #45, it is possible that we will do MCIP #45 as originally proposed
+and this proposal -- they are orthogonal, strictly speaking. We could decide to adopt this proposal as a secondary layer of defense against mixing problems, or in service
+of a new consensus feature in the future.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A `network id` is a UTF-8 string. No particular limit on the length is imposed.
```
jgreat, line 84: We have some inconsistent naming around how we document the networks and how we label them in operations.

- prod vs MainNet
- test vs TestNet

We have a lot of hard to change (OK very inconvenient to change) pointers like dns, s3 buckets, grafana and sentry ids that use "prod", "test"... as identifiers. 

Should we make these `network_id` identifiers match them and change our external documentation to match. 

I think it might be confusing to users if they expect MainNet/TestNet and see IDs like "test" and "prod"

garbageslam: One thing I didn't do in this doc yet is propose to standardize any particular network id strings for mainnet or testnet.

If you have specific proposals would love to take them.

text/0049-network-ids.md
```diff
@@ -0,0 +1,183 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+* [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md)
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Should there be any limits imposed on the network-id string? Should any UTF-8 string be valid?
```
jgreat, line 170: I suggest DNS compatible limits. Roughly 63 character limit and a match for `[a-z0-9-]`

jgreat (commented):
I think making the networks aware of who they are in a human readable way is a good way to reduces client and operational support burden. 

Clear errors like "Service network_id doesn't match expected network_id" would be really helpful.

text/0049-network-ids.md
garbageslam: thanks, will do

text/0049-network-ids.md
```diff
@@ -0,0 +1,183 @@
...
+configuration across networks, to try to prevent mixing of test and prod when this endpoint is called.
+
+This fragmentation is harmful for several reasons.
+
+* Increases the complexity of properly configuring / deploying a network, which is already high.
+* The intent of "GOOG" vs. "TESTGOOG" is that it prevents a user from being confused and thinking test data is production data,
+  because they'll see the word "TEST", but testing isn't harmed much because the software otherwise works the same.
+  However, for our software, lacking MCIP #40 or any alternative proposal, client software needs to obtain metadata for each token
+  id it encounters, or it cannot work properly. If this metadata doesn't match for GOOG and TESTGOOG then we may not properly be testing
+  whether GOOG works properly for prod when.
```
wjuan-mob, line 53:
```suggestion
  whether GOOG works properly for prod when GOOG behaviour differs from TESTGOOG.
```

garbageslam: fixed separately, thanks

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+A "network id" is a string. The string must be a valid DNS label:
+
+* Up to 63 bytes long
+* Valid characters are `a-z`, `0-9`, and `-` (hyphen).
+* You can't specify a hyphen at the beginning or end of a label.
+
+Consensus servers are passed this string on startup by either a flag `--network-id` or an environment variable `MC_NETWORK_ID`.
+* It is an error if it is missing.
```
voloshyn, line 91: is => throws?
It throws an error if the `--network-id` is missing

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+* This is called the "configured network id".
+
+Client facing APIs are extended to include an OPTIONAL "expected_network_id" argument.
```
voloshyn, line 94: should `expected_network_id` be a `header` field? Adding it to all APIs seems like a lot of changes.

voloshyn (approved):
I like the idea: 
1. the overhead on the SDKs will be minimal;
2. we don't have to come up with and maintain separate data testssets for separate networks;
3. does not require token range reservation, so if someone wants to mint a token in hexspeak like `0xABADCAFEDEADBEEF` they can;
If `network-id` is a header field, adding it to Fog seems simple enough too, I think it makes sense to add it from the get-go.

text/0049-network-ids.md
garbageslam: I guess my writing style in the past has been to avoid language-specific terms like "throws" (C doesn't throw errors, for Python people it is "raises" ?), and just specify what scenarios are erroneous, at least when writing a spec doc

text/0049-network-ids.md
garbageslam: the only reason I can think not to do it that way is that a header can easily be ignored by the server. but I suppose that protobuf argument can be too

garbageslam: i'm happy for it to be grpc / http header

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+
+The primary alternative is, do nothing:
+
+* Continue to use ad-hoc configuration differences like attestation trust roots and token ids as the primary means to distinguish dev, test, prod environments.
+* Continue to introduce more such ad-hoc configuration differences as more features are added.
+* Cross-chain bridges primarily lean on attestation roots to avoid e.g. accidentally minting on the wrong network, but it is a bit harder to manage the configuration,
+  because attestation roots live in files.
+
+Another alternative: The Ethereum [JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) returns the network id as a string, but in practice all the ids are actually integers.
+The mapping from network names to ids is available at chainlist.org.
```
nick-mobilecoin, line 148:
```suggestion
The mapping from network names to ids is available at [chainlist.org](https://chainlist.org/).
```

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+* Size on the wire is likely not a serious concern. But if it were, another approach would be to send a hash of the network id string, so that we can ensure that the on-the-wire size is bounded,
+  and the error message in case of a mismatch sends back the full string. OTOH it seems that in real life the hash would probably not be shorter than the string network id.
+
+Ethereum actually has two ids like this: A network id, and a chain id.
+
+* Peer-to-peer communication between nodes uses the network id
+* Transaction signatures use the chain id
+
+For most networks the two values are the same.
+Chain ID was introduced in EIP-155 when the Ethereum Classic fork occurred, to prevent mixing of these two networks.
```
nick-mobilecoin, line 161:
```suggestion
Chain ID was introduced in [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md) when the Ethereum Classic fork occurred, to prevent mixing of these two networks.
```

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+and this proposal -- they are orthogonal, strictly speaking. We could decide to adopt this proposal as a secondary layer of defense against mixing problems, or in service
+of a new consensus feature in the future.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A "network id" is a string. The string must be a valid DNS label:
+
+* Up to 63 bytes long
+* Valid characters are `a-z`, `0-9`, and `-` (hyphen).
```
nick-mobilecoin, line 87: Just making sure this is intentionally leaving out `A-Z`.

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The primary alternative is, do nothing:
+
+* Continue to use ad-hoc configuration differences like attestation trust roots and token ids as the primary means to distinguish dev, test, prod environments.
+* Continue to introduce more such ad-hoc configuration differences as more features are added.
+* Cross-chain bridges primarily lean on attestation roots to avoid e.g. accidentally minting on the wrong network, but it is a bit harder to manage the configuration,
+  because attestation roots live in files.
+
+Another alternative: The Ethereum [JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) returns the network id as a string, but in practice all the ids are actually integers.
```
nick-mobilecoin, line 147: Do we have any info on why they ended up sticking with integers?
My main thought is to reduce usage errors i.e. `my-network` vs `mynetwork`, I'm guessing we aren't concerned with that use/misuse.

nick-mobilecoin (approved):

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+- Start Date: 2022-07-25
+- MCIP PR: [mobilecoinfoundation/mcips#0049](https://github.com/mobilecoinfoundation/mcips/pull/0049)
+- Tracking issue: None Yet
+
+# Summary
+[summary]: #summary
+
+Each running mobilecoin consensus network will be started with a string, the "network id".
+
+Clients which make requests to consensus MAY pass the expected network id with their request.
```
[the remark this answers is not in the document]
garbageslam: thank you!

text/0049-network-ids.md
```diff
@@ -0,0 +1,191 @@
...
+If this doesn't match the server, an error is returned, which includes the server-side network id in the response.
+
+# Motivation
+[motivation]: #motivation
+
+Two important and competing engineering goals for all financial software are:
+
+* Enable Rigorous Testing: Properly test the code before it goes live in a test environment that matches prod as closely as possible.
+* Prevent Mixing of Test and Prod: Make it hard for e.g. users to think they are talking to testnet when they are actually talking to production and vice versa.
+
```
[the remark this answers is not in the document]
garbageslam: This was something that Jason surfaced: in a bunch of infra and related docs, we call it "prod" and all the dns and sentry labels contain the string "prod". But in some other contexts and docu we call it "mainnet".

I personally am fine with either string becoming the network id

Right now, "testnet" usually refers to a specific testnet that mobilecoin foundation is running. 
The other networks that we create locally etc. are referred to as "dev networks". The main difference is, the enclave signing keys for dev nets don't have humans in the loop, they are created automatically on an as needed basis and generally don't live off-line somewhere. The testnet has a real off-line signing process just like mainnet, so that making a testnet release involves all the same steps as making a mainnet release.

text/0049-network-ids.md
garbageslam: that is a good point

would be pretty lame if there were like "Mobi1eCoin" net or something

eranrund:
This seems reasonable to me. I do think the network id should make it into the enclave. We don't have to worry about releasing a new enclave immediately but there's no reason to not include this in a future release.
I also think we should add it to fog - I haven't seen a convincing argument why only part of the network services (e.g. consensus) should have this but not others (e.g. fog). It's not a particularly complicated change, and it is backwards-compatible for clients.

Thanks for writing this!

garbageslam:
Yeah I was trying to just control the scope of the change, but in POC work it seems like it's pretty easy to add it to fog also. I will update the proposal.

I think a second step should be that:

* network id makes it into the enclave
* network id is part of the transaction and under the signature, and checked by the enclave
* James proposed (offline) that network id is in the block header. this will give mobilecoind and fog-ingest visibility into it
* James also thinks we should call it "chain id" and we shouldn't have two separate ids if there's no good reason.

Right now I'm just trying to do the network-id part because it can ship without an enclave upgrade

eranrund:
I like those suggestions, I think the MCIP should cover the whole desirable change, we can implement at whatever pace we like.

eranrund (approved):

text/0049-chain-ids.md
```diff
@@ -0,0 +1,234 @@
...
+They expect chain-ids to match a pattern `[-a-zA-Z0-9]{3,47}`, otherwise they replace those chain-ids with a hash.
+This resembles the rules we proposed except that we are case-sensitive and only lower-case letters are allowed.
+
+# Prior art
+[prior-art]: #prior-art
+
+* [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md)
+* [Cosmos chain registry](https://github.com/cosmos/chain-registry)
+* [Cosmos chain id best practices](https://github.com/cosmos/cosmos-sdk/issues/5363)
+
```
sugargoat, line 223: You could also reference that clients can incorporate the chain ID into their BIP44 key derivation schemes - as the example shows for Bitcoin TestNet

https://en.bitcoin.it/wiki/BIP_0044#Examples

sugargoat: But I think in that case the chain id needs to be an int

text/0049-chain-ids.md
```diff
@@ -0,0 +1,234 @@
...
+  to the user of the app, so that they will know exactly how it will appear in real life in prod when they are testing in testnet.
+  This completely defeats the purpose of "separate token ids per network", since the user will no longer be able to distinguish being on prod vs. being on testnet.
+
+We propose to reduce complexity here and separate these concerns.
+
+* "Chain id"-aware clients is a common pattern used by other blockchains to help clients (and users) be aware of what network they are talking to.
+* Chain id can be easily displayed to the user if desired, which is something that none of our clients actually do today.
+   * For example, when using Metamask, a user can select whether to talk to mainnet, an L2 network, or a test network, via a dropdown menu, and see what network they are currently talking to here.
+     Even if we don't support dynamic switching of networks in a particular client, we can still display the network id somewhere, so that the users can discover TEST vs. PROD easily, without
```
remoun, line 71: s/network id/chain ID/ throughout this PR

text/0049-chain-ids.md
```diff
@@ -0,0 +1,234 @@
...
+* URLs of any other chain services
+* Chain ids of any other networks
+* Contract ids for any relevant contracts
+
+This is better than each of these being independent configuration parameters to the software. They should all be changing in lock-step.
+
+In block-version 3, chain-id will be incorporated to consensus in a more comprehensive way:
+
+* The enclave will become `chain-id` aware, so that the network cannot peer if the chain-id's don't match.
+* A `chain_id` field will be added to the block header in block-version 3.
```
remoun, line 120: Nit: Inconsistent usage of `chain-id` vs `chain_id` (- vs _)

remoun (approved):

garbageslam:
sorry guys, this wasn't supposed to merge yet, we were supposed to wait for final comment period, but i fat-fingered it

Please, let's move any further comments here:

https://github.com/mobilecoinfoundation/mcips/pull/51

PR #57, opened 2022-11-03
Proposal to update mixin uniqueness rules for SCIs

garbageslam:
[Rendered Proposal](https://github.com/mobilecoinfoundation/mcips/blob/update-mixin-uniqueness-rules-for-scis/text/0057-update-mixin-uniqueness-rules-for-scis.md)

### DIFF
--- a/text/0057-update-mixin-uniqueness-rules-for-scis.md
+++ b/text/0057-update-mixin-uniqueness-rules-for-scis.md
@@ -132,12 +132,22 @@ of rings.
   collisions. Testing by enforcing hard rules at the level of the consensus network
   is insufficient to ensure security and correctness of the implementation.
 
+Another concern which was raised in review is that the stauts quo allows an active attack along
+the following lines:
+
+* Adversary creates an SCI, selecting as a mixin a TxOut that is conjectured to belong to the victim.
+* Victim tries to match against this SCI, building a transaction using it
+* Adversary observes the `DuplicateRingElement` error and gets nontrivial statistical insight into the
+  victim's TxOut set.
+
+This proposal mitigates this concern.
+
 # Prior art
 [prior-art]: #prior-art
 
 [MCIP 17](https://github.com/mobilecoinfoundation/mcips/pull/17) discussed alternative ring selection strategies.
 
-To our knowledge, no other block chains using ring signatures have a feature like MCIP 31,
+To our knowledge, no other blockchains using ring signatures have a feature like MCIP 31,
 so we don't expect to find prior art around this.
 
 # Unresolved questions
--- a/text/0057-update-mixin-uniqueness-rules-for-scis.md
+++ b/text/0057-update-mixin-uniqueness-rules-for-scis.md
@@ -101,21 +101,26 @@ This has some drawbacks:
 * Harms the user experience, which now requires an additional transaction fee and waiting time
   to accomplish what they were trying to accomplish.
 * Increases the complexity of client code
-* Not much different in terms of ring analysis. If an adversary who can see the rings sees that
-  * An SCI was published and a particular TxOut was one of its ring elements
-  * A Tx was subsequently submitted using that same TxOut as one of its ring elements
-  * One of the outputs of that Tx was subsequently used immediately in a Tx that also consumed this SCI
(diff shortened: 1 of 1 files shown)


eranrund (approved):
This looks good to me - I agree with the assumption that the client can be in charge of ensuring their rings are unique, and that without this change we could end with SCIs that cannot be used by certain users without them doing a self-payment.

remoun (approved):

text/0057-update-mixin-uniqueness-rules-for-scis.md
```diff
@@ -0,0 +1,146 @@
...
+The rationale for this rule is
+* It is generally possible for clients to sample the mixin ring elements without replacement,
+  removing all the real inputs, and so ensure that duplicate ring elements do not occur.
+* Doing so may make it more difficult for an adversary to statistically analyze the rings.
+
+This only justifies why this rule should be a **best practice for clients**, but not why
+it should be enforced at the level of the consensus network. In fact, this rule is not needed
+to protect the integrity of the cryptocurrency -- it doesn't allow double spends or imbalanced
+transactions to remove this rule. It simply helps to prevent clients from using bad ring sampling
+code in production.
```
remoun, line 28:
```suggestion
to protect the integrity of the cryptocurrency -- removing it doesn't allow double spends
or imbalanced transactions. It simply helps to prevent clients from using bad ring sampling
code in production.
```

remoun (approved):

text/0057-update-mixin-uniqueness-rules-for-scis.md
```diff
@@ -0,0 +1,146 @@
...
+or imbalanced transactions. It simply helps to prevent clients from using bad ring sampling
+code in production.
+
+After MCIP #31, the rationale for the rule is no longer universally applicable, because
+there may be multiple parties involved in building a transaction. Consider the following scenario:
+
+1. Alice creates a Signed Contingent Input. Alice's client must choose a ring of mixins at this time.
+   By chance, Alice chooses one of Bob's outputs as a mixin.
+1. Bob tries to create a transaction that incorporates this SCI. Bob's client adds this SCI to the
+   transaction builder, and to pay for it, performs ring selection, also adding some of Bob's outputs.
```
garbageslam, line 36:
```suggestion
   transaction builder, and to pay for it, performs input selection, also adding some of Bob's outputs.
```

text/0057-update-mixin-uniqueness-rules-for-scis.md
```diff
@@ -0,0 +1,146 @@
...
+
+We take the point of view that, the ground has shifted and the rationale for the original consensus rule has been
+undermined. It is no longer the case that
+
+> It is generally possible for clients to sample the mixin ring elements without replacement,
+> removing all the real inputs, and so ensure that duplicate ring elements do not occur.
+
+We propose that, the original rule applies to any inputs without rules, and a lesser rule
+applies to any inputs with rules -- we only check that ring elements are unique within a ring,
+when the input contains rules. This prevents a regression compared to block version 2.
```
garbageslam, line 58:
```suggestion
when the input contains rules. 

SCI's are new in block version 3, which is not yet released. This change does not allow any transactions that would have been invalid in block version 2 to be valid -- it only changes how this historical rule will apply to new transactions that involve block version 3 features.
```

isis-mc (approved):
Interestingly, on the note of the "what can an adversary determine" section, would it be possible to argue that implementing this proposal *strengthens* security because the adversary can no longer e.g. purposefully spam Txs to see which might overlap with an SCI and produce a DuplicateRingElements error?

garbageslam:
> Interestingly, on the note of the "what can an adversary determine" section, would it be possible to argue that implementing this proposal strengthens security because the adversary can no longer e.g. purposefully spam Txs to see which might overlap with an SCI and produce a DuplicateRingElements error?

We could write something about that -- I think it's reversed though:

* The attacker creates an SCI, using a TxOut as a mixin which they think might belong to the counterparty
* The counterparty selects the SCI for a transaction
* The attacker watches to see if `DuplicateRingElement` error occurs, and if it does, the attacker gains some nontrivial information about the likelihood that their guess was right

PR #58, opened 2022-11-09
MCIP Draft - View Authenticated Sender Memo

briancorbin:
[Rendered Proposal](https://github.com/mobilecoinfoundation/mcips/blob/view-authenticated-sender-memo/text/0058-view-authenticated-sender-memos.md)

### DIFF
--- a/text/0058-view-authenticated-sender-memos.md
+++ b/text/0058-view-authenticated-sender-memos.md
@@ -1,6 +1,6 @@
 - Feature Name: `View Authenticate Sender Memo`
 - Start Date: `2022-11-09`
-- MCIP PR: [mobilecoinfoundation/mcips#0055](https://github.com/mobilecoinfoundation/mcips/pull/55)
+- MCIP PR: [mobilecoinfoundation/mcips#0058](https://github.com/mobilecoinfoundation/mcips/pull/58)
 
 # Summary
 [summary]: #summary


text/0058-view-authenticated-sender-memos.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This may potentially reduce the security of the SenderMemo by requiring the view private key (which would potentially be available on an online machine) instead of the spend private key (which is only contained on an offline machine of hardware wallet). If the view private key were to ever be compromised, then the attacker could authenticate their payments using the ViewAuthenticatedSenderMemo of the compromised party, making it appear as if their sent transaction came from the compromised user.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+An alternative would be to generate the SenderMemoCredential during signing and modify the TxOut during the signing process. This would require more discovery as it is uncertain if this is possible without regenerating all of the contents of the TxOut and TxPrefix, which may not be desireable (especially if signing using a hardware wallet).
```
garbageslam, line 34: We currently have designed the hardware wallets stuff to have one round where it computes memo hmacs for these sender memos (using the spend private key), and then a second round where it actually signs ring mlsags.

text/0058-view-authenticated-sender-memos.md
```diff
@@ -0,0 +1,46 @@
...
+- View Authenticated Sender
+- View Authenticated Sender With Payment Request Id Memo
+- View Authenticated Sender With Payment Intent Id Memo
+
+This feature aims to extend the SenderMemo to allow for senders to use their subaddress view private keys to generate Authenticated Sender Memos
+
+# Motivation
+[motivation]: #motivation
+
+With the introduction of the ViewAccountKey and UnsignedTransactions, generating a SenderMemo for RTH became impossible because of the requirement of the Subaddress Spend Private Key during the creation of the TxOuts for the transaction.
```
garbageslam, line 19: I don't think it's actually impossible now, we have a plan here: https://github.com/mobilecoinfoundation/mobilecoin/pull/2585

The idea is to refactor the `MemoBuilder` object implementation that currently requires the spend key to make it more abstract, and allow that to be a hardware device connection instead.

The only one who is impacted here is Copper's multiparty compute flow.

text/0058-view-authenticated-sender-memos.md
garbageslam: If copper really wants us to provide just one payload, and they would have a round where they do memo signing and then another round where they do transaction signing, and those two rounds happen in their application, that seems like it should be possible, it will just make the `UnsignedTx` object a lot messier.

text/0058-view-authenticated-sender-memos.md
```diff
@@ -0,0 +1,46 @@
...
+[drawbacks]: #drawbacks
+
+This may potentially reduce the security of the SenderMemo by requiring the view private key (which would potentially be available on an online machine) instead of the spend private key (which is only contained on an offline machine of hardware wallet). If the view private key were to ever be compromised, then the attacker could authenticate their payments using the ViewAuthenticatedSenderMemo of the compromised party, making it appear as if their sent transaction came from the compromised user.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+An alternative would be to generate the SenderMemoCredential during signing and modify the TxOut during the signing process. This would require more discovery as it is uncertain if this is possible without regenerating all of the contents of the TxOut and TxPrefix, which may not be desireable (especially if signing using a hardware wallet).
+
+The rationale for adding the ViewAuthenticateSenderMemo is that it allows users to utilize RTH with an offline transaction signer or hardware wallet.
```
garbageslam, line 36: The other main alternative is, do nothing. Then, RTH cannot be used with Copper (at least, the authenticated sender memos cannot be).

This may be acceptable -- after all, we did not prioritize making RTH work with mobilecoind, and I'm not sure it works with full-service right now? You would have to tell me.

briancorbin: It works with full service

text/0058-view-authenticated-sender-memos.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+> - Is it possible to modify just the encrypted memo portion of a TxOut without requiring the TxPrefix and OutputSecrets to be regenerated?
```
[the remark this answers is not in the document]
briancorbin: Okay, then that might actually be a plausible solution for offline transaction signing

garbageslam:
So the main thing that makes me think twice about this is, as you say, it reduces the security from the recipient's point of view -- when validating these memos, the signer could be someone who stole the view key (rather than needing to be someone who has the spend key).

If I'm a client and I receive a TxOut with a view authenticated sender memo, do I validate it and then display it in basically the same way as the existing sender memo?

I wonder if there's a way to make it so that clients only trust view authenticated sender memos when it's necessary. So for instance, for a copper-style multisig account, maybe the public address has a flag that says to accept view authenticated sender memos? And then if I'm a client, I only respect view authenticated sender memos if the public address contains that flag? And other "normal" account types aren't affected, and the client needs to see a spend-key authenticated sender memo in order to treat the origin of the payment as verified.

This would prevent any regression around what happens if I'm a "normal" user and someone steals my view key, which today is bad, but doesn't enable forging these authenticated memos.

That's my 2 cents for now, will think some more

PR #59, opened 2022-12-02
Token metadata document

cbeck88:
Standardize a token metadata document and its authentication

[Rendered Proposal](https://github.com/cbeck88/mcips/blob/token-metadata-document/text/0059-token-metadata-document.md)


text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
+- Feature Name: token_metadata_document
+- Start Date: 2022-12-02
+- MCIP PR: [mobilecoinfoundation/mcips#0059](https://github.com/mobilecoinfoundation/mcips/pull/0059)
+- Tracking Issue: TBD
+
+# Summary
+[summary]: #summary
+
+This MCIP specifies a means for clients to dynamically discover metadata about token ids
+on the mobilecoin blockchain, so that they can then display the balance of a new token to customers
```
eranrund, line 10:
```suggestion
on the MobileCoin blockchain, so that they can then display the balance of a new token to customers
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+in an appropriate way.
+
+This proposal differs from an earlier proposal MCIP 40, in that it is a stop-gap mechanism which
```
eranrund, line 13: Consider linking to the MCIP

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+This proposal creates a source of truth for token metadata for tokens on the MobileCoin blockchain.
+
+Two documents are hosted on the MobileCoin foundation website:
+
+* `https://www.mobilecoin.com/token_metadata.json`: A JSON document containing token metadata
+* `https://www.mobilecoin.com/token_metadata.sig`: An ed25519 signature over the bytes of the .json document. This is binary data.
+
+The ed25519 signing key will be controlled by the mobilecoin foundation.
```
eranrund, line 58:
```suggestion
The ed25519 signing key will be controlled by the MobileCoin foundation.
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+This MCIP will be updated, and the public key will be posted here, after the key is created,
+if the proposal is accepted.
+
+Token metadata ed25519 public key hex bytes: `TBD`
```
eranrund, line 62: I wonder if we can use the same key as the minting trust root?

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+
+The `token_metadata.json` document has the following schema (example):
+
+```
+{
+    signing_timestamp: "10020032320032",
+    metadata: [
+        {
+            token_id: "0",
+            currency_name: "Mobilecoin",
```
eranrund:
```suggestion
            currency_name: "MobileCoin",
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+            ticker_symbol: "eUSD",
+            symbol: "$",
+            decimals: 6,
+            logo_svg: "b64...==",
+            info_url: "https://mobilecoin.com/blog/mobilecoin-launches-eusd"
+        },
+        {
+            token_id: "14",
+            currency_name: 'Meowblecoin',
+            ticker_symbol: 'MEOW',
```
eranrund:
```suggestion
            currency_name: "Meowblecoin",
            ticker_symbol: "MEOW",
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+            decimals: 12,
+            logo_svg: "b64...==",
+            info_url: "https://www.meowblecoin.com"
+        }
+    ],
```
eranrund, line 95:
```suggestion
    ]
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+* A `TxOut` with token id of 0 and a (u64) value of `1.5 * 10^12`, may be displayed as "1.5 MOB" to the user, because the token id corresponds to MOB, and the number of decimals of MOB is 12.
+* A `TxOut` with a token id of 1 and a  (u64) value of `5 * 10^8` may be displayed as `500 eUSD` to the user, because the token id corresponds to eUSD, and the number of decimals is 6.
+* That `TxOut` might also be displayed as `$500` because the symbol of `eUSD` is specified as `$`.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The token-metadata json fields have the following semantics:
+
+* `signing_timestamp`: A string representation of the unix timestamp at the time of creating the document's signature.
```
eranrund, line 113: That's not really possible. You're going to be editing this file and then at some point later sign it... maybe this should be `last_updated`?

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+* `token_id`: A decimal string representation of the `u64` corresponding to the token id.
+* `currency_name`: A UTF-8 string which is the full, official name of the currency. This is intended to be an english-language noun or noun phrase.
```
eranrund, line 115:
```suggestion
* `currency_name`: A UTF-8 string which is the full, official name of the currency. This is intended to be an English-language noun or noun phrase.
```

I think English needs to be capitalized

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+* `ticker_symbol`: A ticker symbol for the currency. This is appropriate to be displayed by an exchange, as a short form of the name of the currency. Not more than 12 printable ASCII characters, including `[A-Za-z0-9].-`.
```
eranrund, line 116: I think the right way to express this is `[A-Za-z0-9.-]{1,12}`

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+* As an alternative, we could have a k-of-n ed25519 multisignature. This would allow for the signing authority to be shared across several parties who can sign asynchronously.
+
+Since this signature can easily be performed off-line anyways, and the document isn't expected to change often, we think a single ed25519 signature is fine for this stop-gap proposal.
+
+In our design, we included SVG images inline in the metadata document to avoid this validation issue.
+This will not scale well to e.g. millions of tokens, and we expect to transition to on-chain metadata before that becomes a problem.
+
+This proposal does not envision having different token registries per network. It is believed that this complexity is unnecessary.
+However, it means that just as clients need to hit Intel's IAS services as part of testing (in prod and dev), they will have to hit
+the `mobilecoin.com` URLs as well, so this is another way in which testing a deployed network will not be completely isolated to the servers deployed as part of the network.
```
eranrund, line 157: We could also have the consensus nodes serve this file and signature if we want to.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+We chose json instead of protobuf to make it easier for clients (which may be web browsers) to directly consume this document in e.g. javascript.
```
eranrund, line 162:
```suggestion
We chose JSON instead of protobuf to make it easier for clients (which may be web browsers) to directly consume this document in e.g. Javascript.
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+
+We chose to separate the signature from the document to avoid the complexity of storing serialized json within json.
+This also allows that we can modify the signature scheme in the future while maintaining backwards compatibility, as long as we continue to serve both the old signature and the new signature,
+at different url's.
```
eranrund, line 166:
```suggestion
at different URLs.
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+[prior-art]: #prior-art
+
+We considered the metadata used in both Cardano native assets and Algorand standard assets when considering this proposal.
+We also considered how tokens are assigned metadata in the Ethereum ecosystem.
+
+In the Ethereum ecosystem, most tokens (fungible and non-fungible) have logos, but they are typically stored on centralized servers, and referred to via URLs.
+
+In [EIS-2569](https://eips.ethereum.org/EIPS/eip-2569) it was proposed to actually store SVG data on chain.
+
+Typically, wallets like metamask use services like Infura to access names and logos of tokens. There is not a standard way outside of these
```
eranrund, line 452:
```suggestion
Typically, wallets like MetaMask use services like Infura to access names and logos of tokens. There is not a standard way outside of these
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,485 @@
...
+This MCIP will be updated, and the public key will be posted here, after the key is created,
+if the proposal is accepted.
+
+Token metadata ed25519 public key hex bytes: `TBD`
+
+The `token_metadata.json` document has the following schema (example):
+
+```
+{
+    signing_timestamp: "10020032320032",
```
eranrund: Maybe we should include a schema_version field that is initially set to 1, and incremented when we make changes to the schema/start including other data/etc

eranrund (changes requested):
Nice, thank you. Mostly nits around capitalization and some small suggestions.

jgreat (commented):
Some initial thoughts. 

### How secure does this data need to be? Who needs to verify/approve this data?

To add a token we need foundation approval and signing of the token config. 

Should this metadata just be added to that tokens config document and included as part of the token config signing process?

Should a different party other than the foundation be able to alter or add to the token metadata?

If other parties can alter the token metadata is this a vector for bad actors? Change a token metadata after creation to impersonate another token or brand?

### Client impact and scaling and hosting.

If we scale to hundreds or thousands of tokens a single json file will probably become unwieldy pretty quickly.  Especially if we are embedding svg data inline with the data.

If a client has to download and parse the entire dataset because on a regular basis, this could be problematic over slow or low bandwidth connections.

If this data doesn't need to be accessed over an attested connection, could we add it as an untrusted path or even a different micro-service included under the consensus server hostname?   A service where a client can query for the token ids it needs instead of the entire set?  Is this what mcip 40 proposes? 

I'm not sure www.mobilecoin.com (or mobilecoin.foundation) is a good place for this. We may want to link to this and other mobilecoin network information from the websites, but those paths are generally managed by marketing and or 3rd parties.  

Do we care if the data is centralized (stored in a single place like www.mobilecoin.com) or should it be decentralized, presented and stored by all of the node operators? 

A benefit of keeping it near the consensus service is clients would not need to know about additional paths and or configurations to get this information.

Another solution may be storing this data in the ledger archive s3 compatible buckets. I like this a bit less since it will probably be a pretty manual process for our node operators to keep this data in sync.

text/0059-token-metadata-document.md
cbeck88: That's a good idea, will change the proposal to that

text/0059-token-metadata-document.md
cbeck88: good idea, we can add that too

cbeck88:
> How secure does this data need to be? Who needs to verify/approve this data?
> 
> To add a token we need foundation approval and signing of the token config.
>
> Should this metadata just be added to that tokens config document and included as part of the token config signing process?

I think Eran is right and it could be the same minting trust root. But I would suggest that it should be a separate document, because I think we should separate the data about a token that the enclave needs to function from the metadata that the enclave doesn't care about.

> If other parties can alter the token metadata is this a vector for bad actors? Change a token metadata after creation to impersonate another token or brand?

Yes, this is not something that should be allowed to happen, and one of the design goals is to make it hard for bad actors to do this.

cbeck88:
> Client impact and scaling and hosting.
>
> If we scale to hundreds or thousands of tokens a single json file will probably become unwieldy pretty quickly. Especially if we are embedding svg data inline with the data.
>
> If a client has to download and parse the entire dataset because on a regular basis, this could be problematic over slow or low bandwidth connections.

Yes I agree, this solution isn't expected to be scale if we have more than a few dozen tokens. But we will do something else to improve this before then.

> If this data doesn't need to be accessed over an attested connection, could we add it as an untrusted path or even a different micro-service included under the consensus server hostname? A service where a client can query for the token ids it needs instead of the entire set? Is this what mcip 40 proposes?

The issue with querying just the token ids you need is that you have to be able to authenticate any data that you get, with a chain of trust going back to the foundation. You wouldn't be able to validate the signature with just part of the data. We could change the scheme so that there is a signature over each token id payload, but this will add engineering complexity, and probably isn't needed for now.

cbeck88:
> Do we care if the data is centralized (stored in a single place like [www.mobilecoin.com](http://www.mobilecoin.com/)) or should it be decentralized, presented and stored by all of the node operators?
>
> A benefit of keeping it near the consensus service is clients would not need to know about additional paths and or configurations to get this information.

That seems fine to me -- if all the nodes are using nginx or something as a proxy ahead of the consensus-service, we could simply make nginx serve these static files as well?

cbeck88:
I found some more info about currency localization, will continue to read and assess:

https://learn.microsoft.com/en-us/globalization/locale/currency-formatting
https://polaris.shopify.com/foundations/formatting-localized-currency

text/0059-token-metadata-document.md
nick-mobilecoin: :thought_balloon: Since the `.sig` is a separate file could the signing utility update the value during the signing process?

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+* `decimals`: An integer specifying how the `u64` integer in a TxOut in the blockchain is scaled to compute a user-displayed amount of the currency.
+* `logo_svg`: An optional logo image. This is a base64-encoded SVG document. This is expected to have been sanitized using something like [svg-hush](https://github.com/cloudflare/svg-hush), to remove scripting, hyperlinks to other documents, and references to cross-origin resources. The full extent of such sanitization will not be specified here.
+* `info_url`: A link to a website containing more information about the token that may be interesting to token holders. This should basic information about the purpose of the token, its supply, any utility that it has, or links to associated whitepapers.
+
+Clients MUST download and validate the `token_metadata.sig` before attempting to process the `token_metadata.json`, and must reject the json with an error if the signature fails.
```
nick-mobilecoin, line 139: This may need more detail on _how_.
- Is it only the non whitespace characters being signed, i.e. minified JSON
- Is it the JSON file as presented.  If it's as presented may want to specify '\n' line endings for those that work on windows where '\r\n' is the usual line ending.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+
+Clients MAY continue to bake token metadata into their build at their discretion, for instance, in order to support offline wallets or other cases where the client cannot access the internet. When both baked-in metadata and the `token_metadata.json` are available, the validated `token_metadata.json` should be considered a more reliable source of truth.
+
+Maintainers of this document SHOULD NOT:
+
+* Allow duplicate records for a given token id
+* Delete a token id's record
+* Modify data such as the `ticker_symbol` or `decimals` of a token, which may confuse users, and particularly, any exchanges that use this as a source of truth.
+* There may be legitimate reasons to make changes the `currency_name` or `logo`, but this needs to be carefully considered.
```
nick-mobilecoin, line 150: :upside_down_face: 
(Also here it's `logo` while up above it's `logo_svg`)
```suggestion
* There may be legitimate reasons to make changes to the `currency_name` or `logo`, but this needs to be carefully considered.
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+
+Clients MUST download and validate the `token_metadata.sig` before attempting to process the `token_metadata.json`, and must reject the json with an error if the signature fails.
+
+Clients SHOULD include a minimum for `signing_timestamp` as part of their build, and reject any `token_metadata.json` which is less than the baked-in `signing_timestamp`. This prevents replay attacks where an old `token_metadata.json` document is substituted for the latest one with the goal of confusing the users by making their tokens display differently or not at all. They can fetch the latest `token_metadata.json` at build time to get the latest `signing_timestamp`.
+
+Clients MAY continue to bake token metadata into their build at their discretion, for instance, in order to support offline wallets or other cases where the client cannot access the internet. When both baked-in metadata and the `token_metadata.json` are available, the validated `token_metadata.json` should be considered a more reliable source of truth.
+
+Maintainers of this document SHOULD NOT:
```
nick-mobilecoin, line 145: :wrench: When I first read this line I read "this" as "this MCIP"
```suggestion
Maintainers of the `token_metadata.json` SHOULD NOT:
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+
+* Allow duplicate records for a given token id
+* Delete a token id's record
+* Modify data such as the `ticker_symbol` or `decimals` of a token, which may confuse users, and particularly, any exchanges that use this as a source of truth.
+* There may be legitimate reasons to make changes the `currency_name` or `logo`, but this needs to be carefully considered.
+* Allow two token ids with `currency_name`s which could be confused
+* Allow the same `ticker_symbol` to appear, or, allowing two `ticker_symbols` which differ only in case, or the replacement of letters with similar numbers.
```
nick-mobilecoin, line 152: :question: Can this restriction really be enforced at all with in the `token_metadata.json`?
If an exchange decides to post `M0B` and the foundation decides that it needs to be in the block chain, is there some way to place an alternate symbol here?

```suggestion
* Allow the same `ticker_symbol` to appear, or, allowing two `ticker_symbol`s which differ only in case, or the replacement of letters with similar numbers.
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+We chose not to attempt to standardize the localization of any cryptocurrency names, since it does not seem conventional to do so in cryptocurrency wallets.
+
+We chose to allow numbers and some punctuation to be used in ticker symbols.
+
+This is based on looking at token lists on various cryptocurrency exchanges:
+
+* [Binance.US token list](https://support.binance.us/hc/en-us/articles/360049417674-List-of-Supported-Assets)
+* Kraken's exchange includes the following symbols, where suffixes such as `.S` suffix are used with staked tokens and similar.
+
+```
```
nick-mobilecoin, line 194: Suggest removing the full list and grab a few to highlight like what was done for Coinbase.
This full list feels like a wall of text getting in the way of the more pertinent information.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+>    Creator (required)
+>    AssetName (optional, but recommended)
+>    UnitName (optional, but recommended)
+>    Total (required)
+>    Decimals (required)
+>    DefaultFrozen (required)
+>    URL (optional)
+>    MetaDataHash (optional)
+
+In Cardano, native assets can be created before being [registered with the token registry](https://developers.cardano.org/docs/native-tokens/token-registry/How-to-prepare-an-entry-for-the-registry-NA-policy-script). Only a name is required to create the asset. When subsequently registering an asset, a "description" is required. Then a ticker, url, logo, and decimals are all considered optional. In the Cardano registry, metadata can be deleted and updated, but it requires a signature from the creator.
```
nick-mobilecoin, line 485: :upside_down_face:  nit
The quotes around description felt out of place when ticker, url, logo, etc were naked
```suggestion
In Cardano, native assets can be created before being [registered with the token registry](https://developers.cardano.org/docs/native-tokens/token-registry/How-to-prepare-an-entry-for-the-registry-NA-policy-script). Only a name is required to create the asset. When subsequently registering an asset, a description is required. Then a ticker, url, logo, and decimals are all considered optional. In the Cardano registry, metadata can be deleted and updated, but it requires a signature from the creator.
```

the-real-adammork (approved):

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+tokens that are launched still cannot actually be used by users. This effectively hurts the
+time-to-market on any new token that is created.
+
+There are second order effects to this -- if tokens become "available" in clients at different
+points in time, then users can get bitten in a bad way. For instance, if a Moby user sent eUSD
+to a Signal user, and at the time, Moby supports eUSD but Signal does not, then the recipient
+cannot see the payment in their app, or send it anywhere, or send it back to the Moby user.
+This could lead to disputed payments, bad user experiences, etc.
+
+At the same time, putting token metadata on chain will require an enclave upgrade, while creating
```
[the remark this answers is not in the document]
eranrund: Yes I believe that's what it means

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+We propose that the `MINTING_TRUST_ROOT` key is reused for this purpose.
+
+Minting trust root ed25519 public key hex bytes: `TODO`
+
+The `token_metadata.json` document has the following schema (example):
+
+```
+{
+    "schema_version": 1,
+    "signing_timestamp": "10020032320032",
```
eranrund, line 68: I think this should be changed to `updated_at` or something like that, since the signing happens after this gets changed.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+This is also consistent with [how shopify handles currency localization](https://polaris.shopify.com/foundations/formatting-localized-currency).
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The token-metadata json fields have the following semantics:
+
+* `schema_version`: An integer describing the version number of the schema of this document. This is 1 until a future MCIP proposes schema modifications.
+* `signing_timestamp`: A string representation of the unix timestamp at the time of creating the document's signature.
+* `token_id`: A decimal string representation of the `u64` corresponding to the token id.
```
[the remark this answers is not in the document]
eranrund: Correct. I think its 2**53 according to the spec 😞

text/0059-token-metadata-document.md
eranrund: I think the signature is over whatever bytes are inside the json file. Whether it has `\n` or `\r\n` in it for line endings should not matter.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+
+This is also consistent with [how shopify handles currency localization](https://polaris.shopify.com/foundations/formatting-localized-currency).
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The token-metadata json fields have the following semantics:
+
+* `schema_version`: An integer describing the version number of the schema of this document. This is 1 until a future MCIP proposes schema modifications.
+* `signing_timestamp`: A string representation of the unix timestamp at the time of creating the document's signature.
```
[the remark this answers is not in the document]
cbeck88: My intent was seconds, will update

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,504 @@
...
+* `ticker_symbol`: A ticker symbol for the currency. This is appropriate to be displayed by an exchange, as a short form of the name of the currency. Not more than 12 printable ASCII characters, including `[A-Za-z0-9].-`.
+* `symbol`: Most fiat currencies have a printable character such as $, £, ¥. Some cryptocurrencies do also, Bitcoin has ₿, and Ethereum has Ξ. This may be optionally specified as a UTF-8 string in the "symbol" field. How exactly it is displayed may be locale specific and we do not attempt to formally specify that at this time.
+* `decimals`: An integer specifying how the `u64` integer in a TxOut in the blockchain is scaled to compute a user-displayed amount of the currency.
+* `logo_svg`: An optional logo image. This is a base64-encoded SVG document. This is expected to have been sanitized using something like [svg-hush](https://github.com/cloudflare/svg-hush), to remove scripting, hyperlinks to other documents, and references to cross-origin resources. The full extent of such sanitization will not be specified here.
```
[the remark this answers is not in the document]
cbeck88: I think it's better to leave it as logo_svg, it will help readers of code using this to understand what form the logo takes. Its also possible we will revisit logo formatting and end up with several optional logo fields.

text/0059-token-metadata-document.md
cbeck88: We could rename from "ticker symbol" to "short code", that matches shopify docu around fiat currencies.

My expectation is that whatever we put here will be suggestive of what an exchange uses as ticker symbol (and vice versa).

The thing is, shopify calls it an ISO short code, and I am not sure that cryptocurrencies have ISO short codes the same way fiat currencies do. But maybe this field can just be a non- specific short-code.

text/0059-token-metadata-document.md
nick-mobilecoin: :+1: I think it's clearer.  

There is a chance it fails to validate if a tool replaces line endings or trailing newline during the download process, but I'm not sure how to prevent that without a giant note here that might get in the way.  

It's probably best to wait and see if it's enough of an issue to justify a note or similar in docs.  The worse that happens is someone isn't able to verify the signature.

cbeck88: > There is a chance it fails to validate if a tool replaces line endings or trailing newline during the download process, but I'm not sure how to prevent that without a giant note here that might get in the way.

I don't think this is likely to happen, the document is going to be the body in an HTTP GET response. Curl does not modify this body, and no other standard http library is going to modify the bytes of the body in a way that is unavoidable. Http libraries provide access to the bytes of the body of the HTTP response.

text/0059-token-metadata-document.md
cbeck88: In the consensus nodes, the field is called `signed_at` but I feel like calling it a timestamp is better because it is actually a timestamp, `at` is less specific.

I tried to change the semantics so that this is the "approximate" time of signing. I don't think it's a big deal if the signing happens a bit later than the document is actually changed.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,277 @@
...
+
+* `schema_version`: An integer describing the version number of the schema of this document. This is 1 until a future MCIP proposes schema modifications.
+* `signing_timestamp`: A string representation of the unix timestamp at the (approximate) time of creating the document's signature. This is an integer number of seconds since the unix epoch.
+* `token_id`: A decimal string representation of the `u64` corresponding to the token id.
+* `currency_name`: A UTF-8 string which is the full, official name of the currency. This is intended to be an English language noun or noun phrase.
+* `short_code`: A short code for the currency. This may be displayed to users in a manner similar to ISO 4217 currency codes, for example, USD, GBP, CAD, in connection to an amount of currency. This is expected to match the ticker symbol used for this asset on cryptocurrency exchanges. Not more than 12 printable ASCII characters, including `[A-Za-z0-9].-`.
+* `symbol`: Most fiat currencies have a printable character such as $, £, ¥. Some cryptocurrencies do also. Bitcoin has ₿. Ethereum has Ξ. This may be optionally specified as a UTF-8 string in the "symbol" field.
```
remoun, line 140: Nit: Specify encoding first, for consistency
```suggestion
* `short_code`: A short code for the currency, up to 12 ASCII characters matching `/[A-Za-z0-9].-/`. This may be displayed to users in a manner similar to ISO 4217 currency codes, for example, USD, GBP, CAD, in connection to an amount of currency. This is expected to match the ticker symbol used for this asset on cryptocurrency exchanges.
* `symbol`: Optional UTF-8 string. Most fiat currencies have a printable character such as $, £, ¥. Some cryptocurrencies do also. Bitcoin has ₿. Ethereum has Ξ.
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,277 @@
...
+* `decimals`: An integer specifying how the `u64` integer in a TxOut in the blockchain is scaled to compute a user-displayed amount of the currency. For example, MOB has 12 decimals, which indicates that `10^12` of the smallest representable units of MOB on the MobileCoin blockchain are equal to one MOB.
+* `logo_svg`: An optional logo image. This is a base64-encoded SVG document. This is expected to have been sanitized using something like [svg-hush](https://github.com/cloudflare/svg-hush), to remove scripting, hyperlinks to other documents, and references to cross-origin resources. The full extent of such sanitization will not be specified here.
```
remoun, line 142: On further thought, how about we omit the base64? That would make it easy for humans to verify the SVG has safe elements, as well as being 25% smaller and gzipping better: https://css-tricks.com/probably-dont-base64-svg/

text/0059-token-metadata-document.md
remoun: Or equivalently, have a flag controlling the host used to fetch the token metadata.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,277 @@
...
+| The negative sign before the number but behind the currency symbol | Denmark            | kr-127,54  |
+| The negative sign after the number                                 | Netherlands        | € 127,54-  |
+| Enclosed in parentheses.                                           | US                 | ($127.54)  |
+
+Most currencies use the same decimal and thousands separator that other numbers in the locale use, but this is not always true. In some places in Switzerland, they use the period as a decimal separator for Swiss francs (Sfr. 127.54), but then use commas as the decimal separator everywhere else (127,54).
+
+We take the point of view that these differences are not "token metadata", which are determined by the ID of the token that we are using, but rather localization properties determined by the locale of the user. So, for example the same app might be capable of displaying `kr 127,54` and `127.54 kr`, depending on the locale of the user, and token metadata does not influence this.
+
+This is also consistent with [how shopify handles currency localization](https://polaris.shopify.com/foundations/formatting-localized-currency).
+
```
remoun, line 129: Nit: This shopify link seems out of place here. Since it's listed under Prior Art, how about we remove this
```suggestion
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,277 @@
...
+> These eight parameters can *only* be specified when an asset is created.
+>
+>    Creator (required)
+>    AssetName (optional, but recommended)
+>    UnitName (optional, but recommended)
+>    Total (required)
+>    Decimals (required)
+>    DefaultFrozen (required)
+>    URL (optional)
+>    MetaDataHash (optional)
```
remoun, line 259: This list gets collapsed in the block quote
```suggestion
>   * Creator (required)
>   * AssetName (optional, but recommended)
>   * UnitName (optional, but recommended)
>   * Total (required)
>   * Decimals (required)
>   * DefaultFrozen (required)
>   * URL (optional)
>   * MetaDataHash (optional)
```

remoun (approved):

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,277 @@
...
+MCowBQYDK2VwAyEA6rqMXns4wNN+W16Eblsue+gqeXlW5C5WhN3MGCc1Ntw=
+-----END PUBLIC KEY-----
+```
+
+The `token_metadata.json` document has the following schema (example):
+
+```
+{
+    "schema_version": 1,
+    "signing_timestamp": "10020032320032",
```
remoun, line 74: Let's use a realistic example
```suggestion
    "signing_timestamp": "1670377814",
```

text/0059-token-metadata-document.md
cbeck88: Honestly i feel like we should not do inline SVG, we should just have a url for the SVG and a hash of the SVG data. I think base64 and not base64 are both pretty human (and internet) unfriendly

remoun: Fine by me, as long as we include a hash/signature of the actual image.

stevenspiel: We may want to consider that a client won't always have internet connection, so if the images aren't cached, they won't be available.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,275 @@
...
+    "schema_version": 1,
+    "signing_timestamp": "1670377814",
+    "metadata": [
+        {
+            "token_id": "0",
+            "currency_name": "MobileCoin",
+            "short_code": "MOB",
+            "decimals": 12,
+            "logo_svg": "b64...==",
+            "info_url": "https://www.mobilecoin.com"
```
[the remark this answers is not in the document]
cbeck88: No, we didn't think about that. Happy to include that

I think Kraken has a simliar thing in their API:

https://docs.kraken.com/rest/#tag/Market-Data/operation/getAssetInfo

They have "decimals" and "display_decimals", where the former is used for record keeping and the latter for actual display to users

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,275 @@
...
+* `decimals`: An integer specifying how the `u64` integer in a TxOut in the blockchain is scaled to compute a user-displayed amount of the currency. For example, MOB has 12 decimals, which indicates that `10^12` of the smallest representable units of MOB on the MobileCoin blockchain are equal to one MOB.
+* `logo_svg`: An optional logo image. This is a base64-encoded SVG document. This is expected to have been sanitized using something like [svg-hush](https://github.com/cloudflare/svg-hush), to remove scripting, hyperlinks to other documents, and references to cross-origin resources. The full extent of such sanitization will not be specified here.
+* `info_url`: A link to a website containing more information about the token that may be interesting to token holders. This should have basic information about the purpose of the token, its supply, any utility that it has, or links to associated whitepapers.
+
+Clients MUST download and validate the `token_metadata.sig` against the bytes of the `token_metadata.json` and the `MINTING_TRUST_ROOT` public key before attempting to process the `token_metadata.json`. Clients MUST reject the json with an error if the signature checking fails.
```
david1alvarez, line 143: I expect situations will arise where clients are unaware of this requirement. Creating a helper method that parses the token_metadata.json and returns an error on signature mismatch feels important for meeting this requirement. If the json file is publicly available, it is otherwise likely there will be instances where it is fetched and used raw.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,275 @@
...
+
+Clients SHOULD include a minimum for `signing_timestamp` as part of their build, and reject any `token_metadata.json` which is less than the baked-in `signing_timestamp`. This prevents replay attacks where an old `token_metadata.json` document is substituted for the latest one with the goal of confusing the users by making their tokens display differently or not at all. They can fetch the latest `token_metadata.json` at build time to get the latest `signing_timestamp`.
+
+Clients MAY continue to bake token metadata into their build at their discretion, for instance, in order to support offline wallets or other cases where the client cannot access the internet. When both baked-in metadata and the `token_metadata.json` are available, the validated `token_metadata.json` should be considered a more reliable source of truth.
+
+Maintainers of the `token_metadata.json` SHOULD NOT:
+
+* Allow duplicate records for a given token id
```
david1alvarez, line 151: This may just be a wording issue, but I suggest we allow duplicate records for a given token id to exist across different versions of the token_metadata.json file. Requiring uniqueness within a given token_metadata.json file seems appropriate.

david1alvarez: This would allow for the concept of an active token_metadata.json file, and would prevent users from needing to fetch all token_metadata.json files to collect token data for all tokens.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,275 @@
...
+* It also would take longer to develop since it would require an enclave upgrade.
+
+This proposal's authentication scheme is a single ed25519 signature.
+
+* As an alternative, we could have a signature chain, such as an x509. This allows for more sophisticated strategies around the keeping the key secure.
+* As an alternative, we could have a k-of-n ed25519 multisignature. This would allow for the signing authority to be shared across several parties who can sign asynchronously.
+
+Since this signature can easily be performed off-line anyways, and the document isn't expected to change often, we think a single ed25519 signature is fine for this stop-gap proposal.
+
+In our design, we included SVG images inline in the metadata document to avoid this validation issue.
```
david1alvarez, line 174: may need to update wording based on svg conversation above, off of line 140

david1alvarez (approved):

text/0059-token-metadata-document.md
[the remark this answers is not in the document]
cbeck88: Added per your suggestion

text/0059-token-metadata-document.md
holtzman: came here to discuss having the consensus nodes serve the file as well. Do our nodes already have facilities to serve out files? It would remove a point of central failure for clients to distribute the token metadata to all of the nodes.

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,279 @@
...
+            "currency_name": "MobileCoin",
+            "short_code": "MOB",
+            "decimals": 12,
+            "suggested_precision": 4,
+            "logo_svg": "b64...==",
+            "info_url": "https://www.mobilecoin.com"
+        },
+        {
+            "token_id": "1",
+            "currency_name": "Electronic Dollars",
```
holtzman, line 87:
```suggestion
            "currency_name": "Electronic Dollar",
```

text/0059-token-metadata-document.md
```diff
@@ -0,0 +1,279 @@
...
+* `info_url`: A link to a website containing more information about the token that may be interesting to token holders. This should have basic information about the purpose of the token, its supply, any utility that it has, or links to associated whitepapers.
+
+Clients MUST download and validate the `token_metadata.sig` against the bytes of the `token_metadata.json` and the `MINTING_TRUST_ROOT` public key before attempting to process the `token_metadata.json`. Clients MUST reject the json with an error if the signature checking fails.
+
+Clients SHOULD include a minimum for `signing_timestamp` as part of their build, and reject any `token_metadata.json` which is less than the baked-in `signing_timestamp`. This prevents replay attacks where an old `token_metadata.json` document is substituted for the latest one with the goal of confusing the users by making their tokens display differently or not at all. They can fetch the latest `token_metadata.json` at build time to get the latest `signing_timestamp`.
```
holtzman, line 149: perhaps at build the client should download the current token_metadata.json and its signature, validate it, and include the whole token_metadata,json in the build, not just the signing_timestamp. This way, if at runtime, the downloaded token_metadata.json has an earlier signing_timestamp, the client has something to fall back on.

The best practice could also be to replace the build-time token_metadata.json with a newer, downloaded one, in the apps data store, whenever the fresh fetch passes the signature test and newer-than-cached-version test.

holtzman (approved):

text/0059-token-metadata-document.md
remoun: The consensus nodes don't have to serve the file over HTTP. Since they speak gRPC, we can define an RPC that returns the metadata JSON and signature.

text/0059-token-metadata-document.md
```diff
@@ -195,6 +202,8 @@ We chose not to attempt to standardize the localization of any cryptocurrency na
 
 We chose to allow numbers and some punctuation to be used in ticker symbols.
 
+We chose not to inline the SVG logo images into the document, because it will make the document bigger and more unwieldy, and harder to edit, particularly if the SVG has to be escaped and stored as a string. Base64-encoded SVG is considered an anti-pattern which increases the size of the document and makes it harder to consume. The logos are expected to be online at reachable URLs anyways, so simply storing their blake2b hash in the signed metadata document reduces the size and complexity of the signed metadata document. There are blake2b libraries available in javascriptThis pattern is also extensible if we need to have more than one version of the logo appropriate for different form factors, we can simply add more `logo_svg_size_url...` fields.
```
nick-mobilecoin, line 205:
```suggestion
We chose not to inline the SVG logo images into the document, because it will make the document bigger and more unwieldy, and harder to edit, particularly if the SVG has to be escaped and stored as a string. Base64-encoded SVG is considered an anti-pattern which increases the size of the document and makes it harder to consume. The logos are expected to be online at reachable URLs anyways, so simply storing their blake2b hash in the signed metadata document reduces the size and complexity of the signed metadata document. There are blake2b libraries available in javascript. This pattern is also extensible if we need to have more than one version of the logo appropriate for different form factors, we can simply add more `logo_svg_size_url...` fields.
```

cbeck88:
This is now entering final comment period

cbeck88:
There's some discussion in the background around changing the URLs where the registry is reachable, so im going to remove final comment period label

PR #61, opened 2023-02-03
Defragmentation Memos

dolanbernard:
[Rendered MD](https://github.com/dolanbernard/mcips/blob/defrag-memo/text/0061-defrag-memos.md)


text/0061-defrag-memos.md
```diff
@@ -0,0 +1,68 @@
...
+
+Including a defragmentation ID may also not be necessary. Conversely, it may be better to use a larger value. A single u64 was chosen for simplicity and should be enough for one account. If clients do not need the ID field, they can also choose to ignore it.
+
+It may also be viable to implement this without removing change outputs from defragmentation transactions. The defragmentation memo could just be written to both outputs. Clients could differentiate the `TxOuts` still as the change amount would be 0. Change outputs are not necessary though for defragmentation, so it makes more sense to remove them. They will always have a value of 0 for defragmentation transactions and are not necessary for RTH purposes.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is a defragmentation ID necessary or valuable? Should it be larger than 64 bits?
```
stevenspiel, line 61: I think there's value in a defrag ID which would help group them together

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,68 @@
...
+| 23 - 64    | Unused bytes |
+
+The fee amount is stored as a u56 in the first 7 bytes. This is done for consistency with RTH Destination memos. The next 8 bytes contain the total outlay of the defragmentation transaction. Lastly, the defragmentation ID is stored in 8 bytes as a u64.
+
+This memo will be written to the payload `TxOut` of a defragmentation transaction. Defragmentation transactions should not produce a change `TxOut`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Accounts which have already performed defragmentation will either have no memos or sender/destination memos for their old defragmentation transactions. Though we encourage not using an account with more than one application, not all developers will implement protections against it nor will all users heed this advice. If new applications do not handle legacy defragmentation transactions, they may inadvertently display them as normal transactions.
```
the-real-adammork, line 45: > Though we encourage not using an account with more than one application

This makes sense but where/when did we decide this, and what is the rationale.

the-real-adammork (approved):

text/0061-defrag-memos.md
dolanbernard: Well, if we just think about the difference between current desktop and mobile wallets, if you have MOB sent to your desktop wallet address, a Fog client won't be able to find the TxOut. So that's one incompatibility. But consider also that all clients (desktop or mobile) may want/need to store info on their own back-end to support additional payment app features. MobileCoin provides, for example, memo types for payment requests and intents. But no such thing as a payment intent or request exists on the blockchain. To make use of these features, client applications will need to use some third-party back-end with info explaining the details of those requests/intents. It wouldn't make sense that different clients from different developers would have knowledge of each other's payment requests/intents.

the-real-adammork: The rationale seems reasonable but "Though we" in this context implies MobileCoin has taken a public position, yet in some cases we've recommended people load their Signal entropy into Moby. Its a nit-pick but I don't think we should refer to the authority of company position unless its been made public.

dolanbernard: Yeah, I agree. The word "we" should not have been used here and I will update this paragraph

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,68 @@
...
+
+Including a defragmentation ID may also not be necessary. Conversely, it may be better to use a larger value. A single u64 was chosen for simplicity and should be enough for one account. If clients do not need the ID field, they can also choose to ignore it.
+
+It may also be viable to implement this without removing change outputs from defragmentation transactions. The defragmentation memo could just be written to both outputs. Clients could differentiate the `TxOuts` still as the change amount would be 0. Change outputs are not necessary though for defragmentation, so it makes more sense to remove them. They will always have a value of 0 for defragmentation transactions and are not necessary for RTH purposes.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is a defragmentation ID necessary or valuable? Should it be larger than 64 bits?
+- Does this memo need any sort of validation? If client applications are hiding defragmentation transactions, users with custom clients could create 'invisible' transactions. This doesn't seem like a major issue, however.
```
eranrund, line 62: I think this can be mitigated by using some of the unused bytes to hold a signature that can only be created by the account owner.

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,68 @@
...
+
+This proposal will also require minor modifications to SDK clients. Since the addition of RTH, all transactions will generate a change `TxOut` regardless of whether or not there is change or if RTH memos are even being written at all. It is fairly trivial to make the required changes, however.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+One alternative to defragmentatino memos would be to continue using RTH for identifying defragmentation transactions. But this approach has a higher barrier of entry for client developers in terms of MobileCoin knowledge. It is much simpler to just check the memo type; and when it comes to sharing example code, defragmentation memos are much easier to demonstrate.
```
cbeck88, line 52:
```suggestion
One alternative to defragmentation memos would be to continue using RTH for identifying defragmentation transactions. But this approach has a higher barrier of entry for client developers in terms of MobileCoin knowledge. It is much simpler to just check the memo type; and when it comes to sharing example code, defragmentation memos are much easier to demonstrate.
```

text/0061-defrag-memos.md
cbeck88: I think we should specify that the authentication works the same as for destination memos? that the memo belongs to the change subaddress? since its a self-payment

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,68 @@
...
+
+Including a defragmentation ID may also not be necessary. Conversely, it may be better to use a larger value. A single u64 was chosen for simplicity and should be enough for one account. If clients do not need the ID field, they can also choose to ignore it.
+
+It may also be viable to implement this without removing change outputs from defragmentation transactions. The defragmentation memo could just be written to both outputs. Clients could differentiate the `TxOuts` still as the change amount would be 0. Change outputs are not necessary though for defragmentation, so it makes more sense to remove them. They will always have a value of 0 for defragmentation transactions and are not necessary for RTH purposes.
```
cbeck88, line 56: +1 on removing change outputs here

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,68 @@
...
+
+From the client's perspective, these larger `TxOut`s created by defragmentation appear as `TxOut`s received through normal transactions. This would be confusing to users if displayed this way in an application. Currently, the only way to differentiate defragmentation transactions from regular ones is to use [RTH](https://github.com/mobilecoinfoundation/mcips/pull/4) and check if the sender and recipient are both the user's account. This isn't immediately clear though, especially to new developers who have not been introduced to concepts like fragmentation or [RTH](https://github.com/mobilecoinfoundation/mcips/pull/4). Thus, a new memo type is proposed to mark defragmentatino transactions.
```
cbeck88, line 15:
```suggestion
From the client's perspective, these larger `TxOut`s created by defragmentation appear as `TxOut`s received through normal transactions. This would be confusing to users if displayed this way in an application. Currently, the only way to differentiate defragmentation transactions from regular ones is to use [RTH](https://github.com/mobilecoinfoundation/mcips/pull/4) and check if the sender and recipient are both the user's account. This isn't immediately clear though, especially to new developers who have not been introduced to concepts like fragmentation or [RTH](https://github.com/mobilecoinfoundation/mcips/pull/4). Thus, a new memo type is proposed to mark defragmentation transactions.
```

cbeck88 (approved):
looks good to me -- i think we should specify that these self-payments also go to the change subaddress though, and this will resolve your question about authentication

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,68 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This proposal adds one new encrypted memo type, the Defragmentation Memo. Defragmentation transactions will write this memo to the payload `TxOut`. Defragmentation transactions will also not produce a change `TxOut`, as there should never be change and it is not needed to construct RTH. This memo has type bytes `0x0002`.
```
cbeck88, line 20: So, part of the reason that we currently write a change output, even if it's value is zero, is that it makes more of your transactions indistinguishable.

With this MCIP, if I see you submit a transaction and only one output is produced, I know that you are submitting a self-payment.

It's better not to reveal that if we don't have to. And the cost of writing an additional zero value output to the chain is pretty low. So I'd suggest to still write the change output anyways.

cbeck88 (changes requested):
We should still write a zero value change output, even for defragmentation transactions, IMO

cbeck88 (approved):
looks good to me, thanks for thinking through and documenting about the second TxOut here

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Summary
+[summary]: #summary
+
+This proposal introduces a new memo type to support [Recoverable Transaction History (RTH)](https://github.com/mobilecoinfoundation/mcips/pull/4). The new memo type will be used by clients to mark `TxOut`s created as part of account defragmentation. This can be used by client applications as a reliable way to exclude defragmentation transactions from appearing as regular transactions.
+
+# Motivation
+[motivation]: #motivation
+
+[RTH](https://github.com/mobilecoinfoundation/mcips/pull/4) introduced a standard method of utilizing [encrypted memos](https://github.com/mobilecoinfoundation/mcips/pull/3) for recording transaction history between two users. Not all transactions on the MobileCoin blockchain are between two different users or accounts, however. Defragmenation transactions are transactions that a user sends to themselves in order to consolidate small `TxOut`s into larger ones. This is necessary if a user needs to send a large transaction but cannot satisfy the `Amount` using 16 `TxOut`s or less.
```
cbeck88, line 13:
```suggestion
[RTH](https://github.com/mobilecoinfoundation/mcips/pull/4) introduced a standard method of utilizing [encrypted memos](https://github.com/mobilecoinfoundation/mcips/pull/3) for recording transaction history between two users. Not all transactions on the MobileCoin blockchain are between two different users or accounts, however. Defragmenation transactions are transactions that a user sends to themselves in order to consolidate small `TxOut`s into larger ones. This is necessary if a user needs to send a large transaction, and has enough balance, but cannot satisfy the `Amount` using their 16 largest `TxOut`s.
```

cbeck88:
can we add a tracking issue, or a link to the implementation PR?

dolanbernard:
> can we add a tracking issue, or a link to the implementation PR?

[Open PR](https://github.com/mobilecoinfoundation/mobilecoin/pull/3170)

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,70 @@
+- Feature Name: RTH Defragmentation Memos
+- Start Date: 2023-02-03
+- MCIP PR: [mobilecoinfoundation/mcips#0061](https://github.com/mobilecoinfoundation/mcips/pull/0061)
+
```
cbeck88, line 4:
```suggestion
- Implementation PR: [mobilecoinfoundation/mobilecoin#3170](https://github.com/mobilecoinfoundation/mobilecoin/pull/3170)

```

voloshyn (approved):

text/0061-defrag-memos.md
briancorbin: So, to expand on this, what we encourage is not using the same account across different Fog Servers, or the absence of one. This is to prevent the situation where a Fog user ends up missing txos. A Desktop Wallet account _can_ be made to support Fog credentials (supported through Full Service), which would make it possible to have the account loaded and useable with no lost information from both instances.

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,71 @@
...
+One alternative to defragmentation memos would be to continue using RTH for identifying defragmentation transactions. But this approach has a higher barrier of entry for client developers in terms of MobileCoin knowledge. It is much simpler to just check the memo type; and when it comes to sharing example code, defragmentation memos are much easier to demonstrate.
+
+Including a defragmentation ID may also not be necessary. Conversely, it may be better to use a larger value. A single u64 was chosen for simplicity and should be enough for one account. If clients do not need the ID field, they can also choose to ignore it.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- ~~Is a defragmentation ID necessary or valuable? Should it be larger than 64 bits?~~
+- ~~Does this memo need any sort of validation? If client applications are hiding defragmentation transactions, users with custom clients could create 'invisible' transactions. This doesn't seem like a major issue, however.~~
+
```
nick-mobilecoin, line 66: :wrench: These are rendered as strikethrough, implying they are resolved. Should they be removed?
```suggestion
None

```

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,71 @@
...
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+This proposal should make it much easier for developers to create apps that behave more like regular payment apps. It will also reduce the amount of code needed to identify defragmentation transactions.
```
nick-mobilecoin, line 70: :wrench: This feels like "rational" not "future possibilities" suggest moving this into the rational portion

text/0061-defrag-memos.md
```diff
@@ -0,0 +1,71 @@
...
+| ---------- | ---- |
+| 0 - 7      | Big-endian bytes of fee `Amount` value, an unsigned 56-bit number |
+| 7 - 15     | Big-endian bytes of the total outlay `Amount` value, an unsigned 64-bit integer |
+| 15 - 23    | Big-endian bytes of the defragmentation ID, an unsigned 64-bit integer |
+| 23 - 64    | Unused bytes |
+
+The fee `Amount` value is stored as a u56 in the first 7 bytes. This is done for consistency with RTH Destination memos. The next 8 bytes contain the total outlay of the defragmentation transaction. Lastly, the defragmentation ID is stored in 8 bytes as a u64.
+
+This memo will be written to both `TxOut`s of a defragmentation transaction. One output will have the non-zero defragmentation `Amount`. The other output will have the change `Amount`, which is always 0 for defragmentation transactions. The non-zero `Amount` output will get a memo with the fee and outlay for that transaction. The 0 `Amount` output will get a memo with a fee and outlay of 0, but the same ID as the non-zero output. Both of the `TxOut`s are sent to the users change subaddress. The memos can be validated by checking that the subaddress the `TxOut`s were sent to is the change subaddress.
```
nick-mobilecoin, line 43: :upside_down_face: 
```suggestion
This memo will be written to both `TxOut`s of a defragmentation transaction. One output will have the non-zero defragmentation `Amount`. The other output will have the change `Amount`, which is always 0 for defragmentation transactions. The non-zero `Amount` output will get a memo with the fee and outlay for that transaction. The 0 `Amount` output will get a memo with a fee and outlay of 0, but the same ID as the non-zero output. Both of the `TxOut`s are sent to the user's change subaddress. The memos can be validated by checking that the subaddress the `TxOut`s were sent to is the change subaddress.
```

nick-mobilecoin (approved):
mainly syntax/display nits

Was there any consideration into standardizing on a defrag sub address?  Not sure a defrag sub address provides much benefit and the memo field seems like it will be more ergonomic for clients, just curious.

text/0061-defrag-memos.md
dolanbernard: Yeah, these are resolved. We're OK to remove

text/0061-defrag-memos.md
dolanbernard: I think you're right, and these points are covered by Motivation already, so I'm fine to remove these.

PR #62, opened 2023-03-02
Recoverable SCIs

briancorbin:


text/0062-recoverable-scis.md
```diff
@@ -0,0 +1,112 @@
...
+
+After an SCI is sent, an app can check if it has been claimed based on if this `TxOut` has been spent. Until it has been spent, the app can potentially cancel the `SCI` by sending a self payment with the `TxOut`.
+
+We additionally extend Recoverable Transaction History memos from [MCIP #4](0004-recoverable-transaction-history.md) to handle `SCI`s as well.
+
+| Memo type bytes | Name                  |
+| --------------- | --------------------- |
+| 0x0301          | SCI generation memo   |
+| 0x0302          | SCI cancellation memo |
+| 0x0003          | SCI sender memo       |
```
wjuan-mob, line 39: Do we need a sender memo? We're imagining these scis are mostly traded around via the deqs, so it is unlikely that the identify of the sci creator is known to the sci executor.

text/0062-recoverable-scis.md
```diff
@@ -0,0 +1,112 @@
...
+# Guide-level explanation - WIP (awaiting feedback and discussion)
+
+[guide-level-explanation]: #guide-level-explanation
+
+When one wishes to generate an SCI, the app must first make a self payment to the `SCI_SUBADDRESS_INDEX = u64::MAX - 3`, which is now reserved for `TxOut`s used as `SCI`s.
+Apps should not select `TxOut`s at this subaddress for inputs in other transactions, and they may choose not to include these in the account balances as they represent the amount of active `SCI`s.
+
+The `TxOut` generated for use as the SCI must contain an `SCI Generation Memo` to identify the details of the SCI it was included in (specifics TBD as feature evolves from feedback and discussion)
+
+After an SCI is sent, an app can check if it has been claimed based on if this `TxOut` has been spent. Until it has been spent, the app can potentially cancel the `SCI` by sending a self payment with the `TxOut`.
```
eranrund, line 31: I think it's worth elaborating on how the app differentiates between canceling an SCI and the SCI being used. In both cases the TxOut will get spent.

text/0062-recoverable-scis.md
eranrund: It's worth mentioning that the consumer of the SCI, who includes a TxOut to pay back the originator change as part of a partial fill, cannot alter the memo, so it is up to the originator to put something there.
There should probably be a memo type that indicates this is a change or partial change - there could be required change input rule (represented as a required output) and/or a partial change input rule.
I'm not sure what information we'd want to include in that memo. Some things we might consider putting there are the public key of the original TxOut, and maybe for partial change the max amount that was available?

eranrund (commented):
Good start, thanks for writing this. I think it needs to lay out very specifically which memo goes into all the following TxOuts involved:
1) The TxOut that goes into the SCI_SUBADDRESS_INDEX
2) The TxOut that cancels an SCI by self-spending an output from the SCI_SUBADDRESS_INDEX
3) Any required outputs - https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/core/src/input_rules.rs#L34
4) Any partial fill outputs - https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/core/src/input_rules.rs#L46
5) Partial fill change output - https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/core/src/input_rules.rs#L60
6) TxOuts generated by the consumer of an SCI to pay themselves whatever they are taking from the SCI

PR #65, opened 2023-05-31
Support verifying DCAP evidence

nick-mobilecoin:
Add an MCIP which describes how the DCAP verification logic should live
side by side with the EPID logic for the clients.

[Rendered MCIP](https://github.com/nick-mobilecoin/mcips/blob/dcap-attestation-evidence/text/0065-dcap-attestation-evidence.md)

### DIFF
--- a/text/0065-dcap-attestation-evidence.md
+++ b/text/0065-dcap-attestation-evidence.md
@@ -1,0 +1,139 @@
+- Feature Name: `dcap_attestation_evidence`
+- Start Date: 2023-05-31
+- MCIP PR: [mcips #65](https://github.com/mobilecoinfoundation/mcips/pull/65)
+- Tracking Issue: [mobilecoin #3119](https://github.com/mobilecoinfoundation/mobilecoin/issues/3119)
+
+# Summary
+[summary]: #summary
+
+Support for verifying both EPID and DCAP quotes in the same codebase.
+
+# Motivation
+[motivation]: #motivation
+
+We want to provide uninterrupted client support of the MobileCoin services when the services transition from EPID to DCAP attestation.
+
+In order to support this uninterrupted transition, clients will need to have
+logic for verifying both the EPID and DCAP flavors of attestation evidence.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For the current EPID attestation the client provides the expected measurements and advisories to a verifier. The verifier then checks the evidence against the expected measurements and advisories.
+
+A _Measurement_ is either a `MRENCLAVE` value or a `MRSIGNER` value set.
+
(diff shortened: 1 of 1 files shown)


cbeck88:
One thing that's not clear to me is, how do I get the "report data" after this change?

This is needed when clients get the fog report: https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/report/resolver/src/lib.rs

https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/fog/ingest/report/src/lib.rs#L30

Does that part all work the same, it's just the validate functions that are changing?

cbeck88:
What is the `Get Enclave Key` function and `Key` in this diagram? that doesn't ring any bells for me. Is that like the key for noise handshake?

nick-mobilecoin:
> How do I get the "report data"

> What is the `Get Enclave Key`

It's probably me using the wrong terminology
Those are one and the same:
- The Ristretto key here, https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/fog/ingest/report/src/lib.rs#L30
- The Ristretto key here https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/fog/api/src/report_parse.rs#L29
- The Ed25519Public here, https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/watcher/src/verification_reports_collector.rs#L115 Though this doesn't verify

cbeck88:
I'd suggest, maybe this arrow should be `Get Report Data`, and then we can clarify that the report data might be a key or might not be.

cbeck88:
The way that stuff is defined is through `EnclaveIdentity` objects like here:

https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/consensus/enclave/impl/src/identity.rs#L4

https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/ingest/enclave/impl/src/identity.rs

The trait is here:

https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/crypto/ake/enclave/src/identity.rs#L10

And it gets pulled in when forming reports here:

https://github.com/mobilecoinfoundation/mobilecoin/blob/2eb944aed82d050be2209fa78a2358d4c095866f/crypto/ake/enclave/src/lib.rs#L646

But I'm not sure we need to get into it in this MCIP, except to say that the enclave decides what values its report data has. DCAP doesn't care about what bytes are in the report data.

cbeck88:
The report data is 64 bytes, and usually the first 32 bytes is the noise X25519 identity for attested connections with clients and peers, and the second key is like the ed25519 block signing key in consensus, and its the ristretto ingress key in fog ingest

eranrund (approved):

PR #68, opened 2023-12-01
Add DCAP evidence to BlockMetadataContents

nick-mobilecoin:
[rendered text](https://github.com/nick-mobilecoin/mcips/blob/nick/dcap-evidence-on-block-metadata/text/0068-dcap-evidence-in-block-metadata.md)

### DIFF
--- a/text/0068-dcap-evidence-in-block-metadata.md
+++ b/text/0068-dcap-evidence-in-block-metadata.md
@@ -1,0 +1,92 @@
+- Feature Name: dcap-evidence-in-block-metadata
+- Start Date: 2023-12-01
+- MCIP PR: [mobilecoinfoundation/mcips#68](https://github.com/mobilecoinfoundation/mcips/pull/68)
+- Tracking Issue: [mobilecoinfoundation/mobilecoin#2373](https://github.com/mobilecoinfoundation/mobilecoin/issues/2373)
+
+# Summary
+[summary]: #summary
+
+Provide DCAP attestation evidence in the block metadata. This is the compliment
+to the Attestation Verification Report (AVR) used in EPID.
+
+# Motivation
+[motivation]: #motivation
+
+EPID SGX hardware is being phased out. The newer SGX machines use DCAP as the
+attestation protocol. This MCIP is to provide the DCAP version of the evidence
+in the block metadata so that chain integrity can be verified for blocks created
+on the newer hardware.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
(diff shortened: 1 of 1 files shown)


text/0068-dcap-evidence-in-block-metadata.md
```diff
@@ -0,0 +1,92 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+The `BlockMetatdataContents` in block versions 4 or greater contain either a
+`DcapEvidence` or a `VerificationReport`.
+(The `BlockMetatdataContents` was added in [MCIP-0043](0043-block-metadata.md)).
+The `DcapEvidence` contains attestation evidence returned by the DCAP
+attestation process. The `VerificationReport` contains the attestatoin evidence
+returned by the EPID attestation process.
+
+In general, block version 4 or greater is expected to only contain
```
eranrund, line 30: The block version is a configuration parameter for consensus. As such, it is possible for us to launch a DCAP consensus with block version 3...
It seems like we'll need to update the consensus node configurations at the same time we are deploying the DCAP release, since it will start including the `dcap_evidence` in `BlockMetadataContents` and clients need to be prepared for that.
As such, we might want to consider pushing timestamps to block version 5 if we decide we don't want too many major changes in one release.

text/0068-dcap-evidence-in-block-metadata.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The `VerificationReport` and `DcapEvidence` were disparate enough that it didn't
+seem feasible to try and convert a `DcapEvidence` into a `VerificationReport`.
+
+While requiring a version bump, this change is able to be backwards compatible
+such that clients only knowing about version 3 can still consume the blocks.
+They will **not** be able to perform verification due to lacking the necessary
```
eranrund: I think it's worth being more explicit here and state that they will not be able to verify the evidence, but they also will not be able to verify the `BlockMetadataContents` signature. I am not sure this makes the change backwards compatible...

nick-mobilecoin: removed the "backwards compatible" wording and, moved the unable to verify wording into the `drawbacks` section.

text/0068-dcap-evidence-in-block-metadata.md
nick-mobilecoin: At this time, due the new HW and significance of DCAP we're moving timestamps out to post 6.0 and thus they won't be on block 4.

